PR #7, opened 2024-04-02
Rewrite binary with go-flags because it is much easier to extend

zimmski:


cmd/eval-symflower-codegen-testing/cmd/command.go
```diff
@@ -0,0 +1,36 @@
...
+	if _, err := parser.AddGroup("Common command options", "", &Command{}); err != nil {
+		log.Fatalf("Could not add arguments group: %+v", err)
+	}
+
+	// Print the help, when there is no active command.
+	parser.SubcommandsOptional = true
+
+	if _, err := parser.Parse(); err != nil {
+		if e, ok := err.(*flags.Error); ok && e.Type == flags.ErrHelp {
+			return
```
bauersimon, line 28: Will this print the help text? Because you explicitly call `WriteHelp` below.

zimmski: It prints the help already in the go-flags code. Don't like that either, but that is how the default help flag works.

PR #8, opened 2024-04-02
Metrics for evaluation and coverage

bauersimon:


evaluate/metrics.go
```diff
@@ -16,31 +17,37 @@ type Metrics struct {
 	Total uint
 	// Executed is the number of benchmarking candidates with successful execution.
 	Executed uint
+
+	// Coverage holds the coverage of the benchmarking candidates.
+	Coverage []float64
```
zimmski, line 22: Do not use float, because it has problems to keep numbers straight. Use uint for metrics.

bauersimon: But coverage reportings from Go have floating point digits. Or should we just cut them off?

zimmski:
Will fix the test while i review

model/llm/llm.go
```diff
@@ -33,7 +33,7 @@ func NewLLMModel(provider provider.QueryProvider, modelIdentifier string) model.
 
 // llmGenerateTestForFilePrompt is the prompt used to query LLMs for test generation.
 var llmGenerateTestForFilePrompt = bytesutil.StringTrimIndentations(`
-	Given the following Go code file, provide a test file for this code.
+	Given the following Go code file "plain.test" with package "native", provide a test file for this code.
```
zimmski, line 36: Commit subject: "TMP"? Temporary? I usually prefix with TODO to make clear it is a TODO commit.

cmd/eval-symflower-codegen-testing/cmd/evaluate.go
```diff
@@ -29,6 +29,8 @@ type Evaluate struct {
 	// TestdataPath determines the testdata path where all repositories reside grouped by languages.
 	TestdataPath string `long:"testdata" description:"Path to the testdata directory where all repositories reside grouped by languages." default:"testdata/"`
 
+	// CSVPath holds the path to the CSV results file.
+	CSVPath string `long:"csv" description:"Path to the CSV results file." default:"evaluation.csv"`
```
zimmski, line 33: Let's group model-configurations and evaluation-configurations separately

zimmski: Also, not that import that it is a CSV file for the field name.

cmd/eval-symflower-codegen-testing/cmd/evaluate.go
```diff
@@ -112,6 +114,7 @@ func (command *Evaluate) Execute(args []string) (err error) {
 	// Evaluating models and languages.
 	log.Printf("Evaluating models and languages")
 	problemsPerModel := map[string][]error{}
+	metricsPerModel := map[string]evaluate.Metrics{}
```
zimmski, line 117: Always sort.

evaluate/metrics.go
zimmski: Lets move it to the TODO. i didn't expect that we are collecting percentage. We should collect the absolute metrics and then do percentage as a summary. Not as a result of the benchmark. That is imprecise the more data we have.

PR #9, opened 2024-04-03
Ability to select repositories in CLI

bauersimon:


cmd/eval-symflower-codegen-testing/cmd/evaluate.go
```diff
@@ -18,6 +19,8 @@ import (
 type Evaluate struct {
 	// Languages determines which language should be used for the evaluation, or empty if all languages should be used.
 	Languages []string `long:"language" description:"Evaluate with this language. By default all languages are used."`
+	// Repositories determines which repository should be used for the evaluation, or empty if all repositories should be used.
+	Repositories []string `long:"repository" description:"Evaluate with this repository. By default all repositories are used."`
```
zimmski, line 23: Sort

cmd/eval-symflower-codegen-testing/cmd/evaluate.go
```diff
@@ -69,10 +77,22 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		for _, modelID := range command.Models {
 			model := model.Models[modelID]
 			language := language.Languages[languageID]
+			languagePath := filepath.Join(command.TestdataPath, language.ID())
 
-			if err := evaluate.EvaluateRepository(model, language, filepath.Join(command.TestdataPath, language.ID(), "plain")); err != nil {
-				log.Fatalf("%+v", err)
+			repositories, err := os.ReadDir(languagePath)
```
zimmski, line 82: Please do not change this part. Look at the comment `Check that models and languages can be evaluated by executing the "plain" repositories.` This execution should always be the "plain" repository, to check if the language and model and ... code are working at all.

bauersimon: Since LLMs are non-deterministic we have a problem here with the benchmarks itself. Sometimes they come up with the correct solution, sometimes they don't. So a) the benchmark itself will be super flaky and b) testing a model with "plain" does not give us anything, i.e. the free Google model sometimes passes in the "plain" repository and sometimes does not.

zimmski: We need "some" test to make sure that execution is working. Since it sometimes works, and sometimes now. It seems to me that it is either a prompt problem or a temperature one.

bauersimon:
Closed in favor of https://github.com/symflower/eval-symflower-codegen-testing/pull/8.

PR #12, opened 2024-04-03
Query available models of openrouter.ai provider

zimmski:


provider/openrouter/openrouter.go
```diff
@@ -29,11 +29,17 @@ func newOpenRouterProvider() provider.Provider {
...
 	config := openai.DefaultConfig(o.token)
 	config.BaseURL = o.baseURL
-	client := openai.NewClientWithConfig(config)
+
+	return openai.NewClientWithConfig(config)
+}
+
+// Query queries the LLM.
+func (o *openRouterProvider) Query(ctx context.Context, modelIdentifier string, promptText string) (response string, err error) {
+	client := o.client()
```
bauersimon: No need to instantiated the client even... we could do `o.client().CreateChatCompletion...` directly.

provider/openrouter/openrouter.go
```diff
@@ -74,21 +74,17 @@ func (p *openRouterProvider) ID() (id string) {
...
-	modelIDs := []string{
-		"google/gemma-7b-it:free",
-		"gryphe/mythomist-7b:free",
-		"mistralai/mistral-7b-instruct:free",
-		"nousresearch/nous-capybara-7b:free",
-		"openrouter/cinematika-7b:free",
-		"undi95/toppy-m-7b:free",
+func (p *openRouterProvider) Models() (models []model.Model, err error) {
+	client := p.client()
+	ms, err := client.ListModels(context.Background())
```
bauersimon, line 79: No need to instantiated the client even... we could do `o.client().CreateChatCompletion...` directly.

provider/openrouter/openrouter.go
zimmski: Prefer a client object for a clean configuration.

provider/openrouter/openrouter.go
zimmski: Same as https://github.com/symflower/eval-symflower-codegen-testing/pull/12#discussion_r1549576437

PR #16, opened 2024-04-03
Improve prompt request by being more explicit

zimmski:


zimmski:
@bauersimon this doesn't work. Let's focus on other parts. Dropping.

PR #33, opened 2024-04-16
Assessment for generation

bauersimon:


zimmski (changes requested):

evaluate/metrics/assessment.go
```diff
@@ -0,0 +1,40 @@
...
+func registerAssessmentKey(key string) AssessmentKey {
+	assessment := AssessmentKey(key)
+	allAssessmentKeys = append(allAssessmentKeys, assessment)
+
+	return assessment
+}
+
+var (
+	// AssessmentKeyNoExplanations means that a model did not produce additional explanations.
+	AssessmentKeyNoExplanations = registerAssessmentKey("no-explanations")
```
zimmski, line 18: - -> `AssessmentKeyNoAdditionalExplanations`
- -> `no-additional-explanations`

because, it is about the "additional". if we would have a prompt where we require explanations... that key would kind of be the same

evaluate/metrics/assessment.go
```diff
@@ -0,0 +1,40 @@
...
+
+func registerAssessmentKey(key string) AssessmentKey {
+	assessment := AssessmentKey(key)
+	allAssessmentKeys = append(allAssessmentKeys, assessment)
+
+	return assessment
+}
+
+var (
+	// AssessmentKeyNoExplanations means that a model did not produce additional explanations.
```
zimmski, line 17: means -> indicates

evaluate/metrics/assessment.go
```diff
@@ -0,0 +1,40 @@
...
+	AssessmentKeyNoExplanations = registerAssessmentKey("no-explanations")
+)
+
+// Assessments holds numerical assessment metrics.
+type Assessments map[AssessmentKey]uint
+
+// Add sums two assessment objects.
+func (a Assessments) Add(o Assessments) Assessments {
```
zimmski, line 25: I think this should be "Merge" or "Union". "Add" seems to me adding to the same object, but we are adding to a new one.

evaluate/metrics/assessment.go
```diff
@@ -0,0 +1,40 @@
...
+
+var (
+	// AssessmentKeyNoExplanations means that a model did not produce additional explanations.
+	AssessmentKeyNoExplanations = registerAssessmentKey("no-explanations")
+)
+
+// Assessments holds numerical assessment metrics.
+type Assessments map[AssessmentKey]uint
+
+// Add sums two assessment objects.
```
zimmski, line 24: -> `merges two assessments into a new assessment and returns it.` ?

evaluate/metrics/assessment.go
```diff
@@ -0,0 +1,40 @@
...
+	if a == nil {
+		a = Assessments{}
+	}
+	if o == nil {
+		o = Assessments{}
+	}
+
+	assessments := map[AssessmentKey]uint{}
+
+	for _, k := range allAssessmentKeys {
```
zimmski, line 35: Make it more generic.

- create new object
- add keys of first
- add keys of second
- return new

evaluate/metrics/metrics.go
```diff
@@ -70,11 +75,17 @@ func (m Metrics) String() string {
...
+	if assessment == nil {
+		assessment = Assessments{}
+	}
+
 	return []string{
 		fmt.Sprintf("%d", m.Total),
 		fmt.Sprintf("%d", m.Executed),
 		fmt.Sprintf("%d", m.Problems),
 		fmt.Sprintf("%.0f", m.AverageCoverage()),
+		fmt.Sprintf("%d", assessment[AssessmentKeyNoExplanations]),
```
zimmski, line 88: Make this generic. Should be a function that counts all.

Also follow-up: all the metrics we have here should be in the assessment at one point. so would be good to print the whole map right away?

evaluate/metrics/metrics.go
```diff
@@ -83,7 +94,7 @@ func FormatStringCSV(metricsPerModel map[string]Metrics) (string, error) {
 	var out strings.Builder
 	csv := csv.NewWriter(&out)
 
-	if err := csv.Write([]string{"model", "files-total", "files-executed", "files-problems", "coverage-statement"}); err != nil {
+	if err := csv.Write([]string{"model", "files-total", "files-executed", "files-problems", "coverage-statement", "no-explanation"}); err != nil {
```
zimmski, line 97: use existing constant, don't copy the key manually.

evaluate/metrics/assessment.go
zimmski: Now that i a have seen the test cases. i am wondering if this should be something like "More content than required" or something like that. What we are looking for is if there is more than just the code fence. If it is more than we asked for.  So maybe "AnswerContainingMoreThanWeAskedFor" or something?

model/llm/llm.go
```diff
@@ -94,20 +95,24 @@ func (m *llm) GenerateTestsForFile(language language.Language, repositoryPath st
...
-	testContent := prompt.ParseResponse(response)
+	assessment, testContent := prompt.ParseResponse(response)
 
 	// TODO Ask the model for the test file name or compute it in a more sophisticated manner.
 	fileExtension := filepath.Ext(filePath)
 	testFilePath := filepath.Join(repositoryPath, strings.TrimSuffix(filePath, fileExtension)+"_test"+fileExtension)
 
-	return os.WriteFile(testFilePath, []byte(testContent), 0644)
+	if err := os.WriteFile(testFilePath, []byte(testContent), 0644); err != nil {
+		return nil, pkgerrors.WithStack(err)
```
zimmski, line 114: Great find!

model/llm/prompt/parse.go
```diff
@@ -13,16 +14,27 @@ var (
...
 // ParseResponse parses code from a model's response.
-func ParseResponse(response string) (code string) {
+func ParseResponse(response string) (assessment metrics.Assessments, code string) {
+	assessment = metrics.Assessments{}
+
 	// Some models produce duplicated code tags, so unify them if needed.
 	response = codeTagDuplicatedRe.ReplaceAllString(response, "```")
 
 	blocks := bytesutil.GuardedBlocks(response, codeTagRe, codeTagRe)
 	if len(blocks) == 0 { // When no code blocks are found, assume that just the code is returned.
```
zimmski, line 24: This and the next comment... sound like block comments. can you make blocks out of the code please.

```
blocks := bytesutil.GuardedBlocks(response, codeTagRe, codeTagRe)

// When no code blocks are found, assume that just the code is returned.
if len(blocks) == 0 { 
...

// Assume the first code block contains the response code fragment.
	block := blocks[0]

```

model/llm/prompt/parse.go
```diff
@@ -13,16 +14,27 @@ var (
 )
 
 // ParseResponse parses code from a model's response.
-func ParseResponse(response string) (code string) {
+func ParseResponse(response string) (assessment metrics.Assessments, code string) {
+	assessment = metrics.Assessments{}
+
 	// Some models produce duplicated code tags, so unify them if needed.
 	response = codeTagDuplicatedRe.ReplaceAllString(response, "```")
```
zimmski, line 21: We slowly are rewording our "Re" to Match internally can you do -> `codeTagDuplicatedMatch`. At this point i am wondering if "Pattern" is better btw.

model/llm/prompt/parse.go
```diff
@@ -13,16 +14,27 @@ var (
...
-		return strings.TrimSpace(response)
+		assessment[metrics.AssessmentKeyNoExplanations] = 1
+
+		return assessment, strings.TrimSpace(response)
 	}
 	// Assume the first code block contains the response code fragment.
 	block := blocks[0]
 
-	return strings.TrimSpace(codeTagRe.ReplaceAllString(block, ""))
+	responseWithoutBlock := strings.Replace(response, block, "", 1)
```
zimmski, line 32: Please comment what this coding block does.

zimmski:
@bauersimon through. looking forward to getting this merged!

evaluate/metrics/assessment.go
bauersimon: If we reuse the existing object, I would call the operation "update" then... because "merge" and "union" are more like operations that have a result.

evaluate/metrics/assessment.go
bauersimon: Then maybe `AssessmentKeyExcessResponse`?

evaluate/metrics/assessment.go
bauersimon: Will not return it as discussed above.

evaluate/metrics/assessment.go
bauersimon: Or then `AssessmentKeyNoExcessResponse` as we reward when there is no additional stuff.

evaluate/metrics/assessment.go
bauersimon: If we always just update the first object, then we cannot handle the case anymore where it is `nil`... So we must ensure everywhere that it is initialized.

zimmski: Yeah. That is what i mean

- Merge combines objects a and b, to return a new c
- Add adds object b to a and does not return

evaluate/metrics/assessment.go
zimmski: Still sum them up. That is why "Add" sounds good to me

zimmski (approved):

zimmski (approved):

PR #37, opened 2024-04-17
Allow to add to existing assessment collection

zimmski:


zimmski:
@bauersimon please review. Did not see tests so i added them. Let's be stricter from now on. Same rules as internally.

evaluate/metrics/assessment.go
```diff
@@ -24,26 +24,30 @@ var (
...
-	}
+// NewAssessments create a new assessment collection.
+func NewAssessments() Assessments {
+	return map[AssessmentKey]uint{}
+}
 
-	assessments := map[AssessmentKey]uint{}
+// Add adds the given assessment collection to the current one.
+func (a Assessments) Add(x Assessments) {
+	for k, v := range x {
```
bauersimon, line 37: If we don't use `allAssessmentKeys` anymore to ensure all keys are present, then we can remove it.

bauersimon (changes requested):

evaluate/metrics/assessment.go
zimmski: For this method, we do not need it. but it is useful for doing the export of the CSV.

Issue #32, opened 2024-04-16
Automatic response categorization


bauersimon:
Currently we only parse the first code fence from the response.

Next step is to automatically extract if there was additional information (i.e. explanatory text) by a model.

---

- [x] response with no error
- [x] response not empty
- [x] response with code
- [x] response with no excess (extra content)
- [x] response with extra code -> moved to https://github.com/symflower/eval-dev-quality/issues/44
- [x] categorization

zimmski:
> @zimmski checking if the response contained extra code is not straightforward... we could rudimentary compare the number of functions we expect with what we get as a response and search for signatures there?

FFS GitHub. why don't you have threads?!?!?!?!

Sounds good. Comparing number of functions: you can use `symflower symbols $filePath` for that.  If there is more than one symbol for plain, right?

bauersimon:
> > @zimmski checking if the response contained extra code is not straightforward... we could rudimentary compare the number of functions we expect with what we get as a response and search for signatures there?
> 
> FFS GitHub. why don't you have threads?!?!?!?!
> 
> Sounds good. Comparing number of functions: you can use `symflower symbols $filePath` for that. If there is more than one symbol for plain, right?

Yeah... I would say we always expect as many top-level functions in the tests as the input file had, right? One test per function for now?

zimmski:
> > > @zimmski checking if the response contained extra code is not straightforward... we could rudimentary compare the number of functions we expect with what we get as a response and search for signatures there?
> > 
> > 
> > FFS GitHub. why don't you have threads?!?!?!?!
> > Sounds good. Comparing number of functions: you can use `symflower symbols $filePath` for that. If there is more than one symbol for plain, right?
> 
> Yeah... I would say we always expect as many top-level functions in the tests as the input file had, right? One test per function for now?

That will change right away with the nest use-case: src/gitlab.nethead.at/symflower/symflower/language/golang/testdata/sanity.go

zimmski:
@bauersimon i think `response with extra code` is done? But the issue lacks the categorization task.

bauersimon:
> @bauersimon i think response with extra code is done?

errr no... we don't know what too much code is... we know what extra text is.

[...]

PR #39, opened 2024-04-17
More assessments based on a model response

bauersimon:
Part of #32

### DIFF
--- a/evaluate/metrics/assessment.go
+++ b/evaluate/metrics/assessment.go
@@ -35,8 +35,6 @@ func RegisterAssessmentKey(key string) AssessmentKey {
 var (
 	// AssessmentKeyFilesExecutes holds the successfully executed files.
 	AssessmentKeyFilesExecuted = RegisterAssessmentKey("files-executed")
-	// AssessmentKeyFilesProblems holds the files with problems.
-	AssessmentKeyFilesProblems = RegisterAssessmentKey("files-problems")
 
 	// AssessmentKeyCoverageStatement counts the cases where 100% coverage was reached.
 	AssessmentKeyCoverageStatement = RegisterAssessmentKey("coverage-statement")
--- a/evaluate/metrics/assessment.go
+++ b/evaluate/metrics/assessment.go
@@ -111,7 +111,7 @@ func FormatStringCSV(metricsPerModel map[string]Assessments) (string, error) {
 	csv := csv.NewWriter(&out)
 
 	if err := csv.Write(csvHeader()); err != nil {
-		return "", err
+		return "", pkgerrors.WithStack(err)
 	}
 	models := maps.Keys(metricsPerModel)
 	sort.Strings(models)
--- a/evaluate/metrics/assessment.go
+++ b/evaluate/metrics/assessment.go
@@ -8,6 +8,8 @@ import (
 
 	pkgerrors "github.com/pkg/errors"
 	"golang.org/x/exp/maps"
+
+	"github.com/symflower/eval-dev-quality/util"
 )
 
 // AssessmentKey defines a key for a numerical key-value assessment pair.
@@ -23,8 +25,9 @@ var (
 // RegisterAssessmentKey registers a new assessment key.
 func RegisterAssessmentKey(key string) AssessmentKey {
 	assessment := AssessmentKey(key)
-	allAssessmentKeys = append(allAssessmentKeys, assessment)
-	allAssessmentKeysStrings = append(allAssessmentKeysStrings, key)
+
+	allAssessmentKeys = util.InsertToSortedSlice(allAssessmentKeys, assessment)
+	allAssessmentKeysStrings = util.InsertToSortedSlice(allAssessmentKeysStrings, key)
 
 	return assessment
 }
--- a/evaluate/metrics/assessment.go
+++ b/evaluate/metrics/assessment.go
@@ -58,6 +58,21 @@ func (a Assessments) Add(x Assessments) {
 	}
 }
 
+// Equal checks if both assessment collections are equal.
+func (a Assessments) Equal(x Assessments) bool {
+	if a == nil || x == nil {
+		return a == nil && x == nil
+	}
+
+	for _, key := range allAssessmentKeys {
+		if a[key] != x[key] {
+			return false
+		}
+	}
+
+	return true
+}
+
 // Merge combines two assessment collections into a new assessment collection and returns the new assessment collection.
 func Merge(a Assessments, b Assessments) (c Assessments) {
 	c = NewAssessments()
--- a/evaluate/metrics/assessment.go
+++ b/evaluate/metrics/assessment.go
@@ -38,8 +38,8 @@ var (
 	// AssessmentKeyCoverageStatement counts the cases where 100% coverage was reached.
 	AssessmentKeyCoverageStatement = RegisterAssessmentKey("coverage-statement")
 
-	// AssessmentKeyNoExcessResponse indicates that a model did not produce more content as requested.
-	AssessmentKeyNoExcessResponse = RegisterAssessmentKey("no-excess-response")
+	// AssessmentKeyResponseNoExcess indicates that a model did not produce more content as requested.
+	AssessmentKeyResponseNoExcess = RegisterAssessmentKey("response-no-excess")
 )
 
 // Assessments holds a collection of numerical assessment metrics.
--- a/evaluate/metrics/assessment.go
+++ b/evaluate/metrics/assessment.go
@@ -39,6 +39,12 @@ var (
 	// AssessmentKeyCoverageStatement counts the cases where 100% coverage was reached.
 	AssessmentKeyCoverageStatement = RegisterAssessmentKey("coverage-statement")
 
+	// AssessmentKeyResponseNoError indicates that a model responded without error.
+	AssessmentKeyResponseNoError = RegisterAssessmentKey("response-no-error")
+	// AssessmentKeyResponseNotEmpty indicates that a model response was not empty.
+	AssessmentKeyResponseNotEmpty = RegisterAssessmentKey("response-not-empty")
+	// AssessmentKeyResponseWithCode indicates that a model responded with code.
+	AssessmentKeyResponseWithCode = RegisterAssessmentKey("response-with-code")
 	// AssessmentKeyResponseNoExcess indicates that a model did not produce more content as requested.
 	AssessmentKeyResponseNoExcess = RegisterAssessmentKey("response-no-excess")
 )
--- a/evaluate/metrics/assessment_test.go
+++ b/evaluate/metrics/assessment_test.go
@@ -134,23 +134,26 @@ func TestAssessmentString(t *testing.T) {
 	}
 
 	validate(t, &testCase{
-		Name: "Initial Metrics",
+		Name: "Empty Metrics",
 
 		Assessment: NewAssessments(),
 
-		ExpectedString: "coverage-statement=0, files-executed=0, response-no-excess=0",
+		ExpectedString: "coverage-statement=0, files-executed=0, response-no-error=0, response-no-excess=0, response-not-empty=0, response-with-code=0",
 	})
 
 	validate(t, &testCase{
-		Name: "Empty Metrics",
+		Name: "Non-empty Metrics",
 
 		Assessment: Assessments{
 			AssessmentKeyCoverageStatement: 1,
 			AssessmentKeyFilesExecuted:     2,
-			AssessmentKeyResponseNoExcess:  4,
+			AssessmentKeyResponseNoError:   4,
+			AssessmentKeyResponseNoExcess:  5,
+			AssessmentKeyResponseNotEmpty:  6,
+			AssessmentKeyResponseWithCode:  7,
 		},
 
-		ExpectedString: "coverage-statement=1, files-executed=2, response-no-excess=4",
+		ExpectedString: "coverage-statement=1, files-executed=2, response-no-error=4, response-no-excess=5, response-not-empty=6, response-with-code=7",
 	})
 }
 
@@ -180,8 +183,8 @@ func TestFormatStringCSV(t *testing.T) {
 		},
 
 		ExpectedString: `
-			model,coverage-statement,files-executed,response-no-excess
-			Model,0,0,0
+			model,coverage-statement,files-executed,response-no-error,response-no-excess,response-not-empty,response-with-code
+			Model,0,0,0,0,0,0
 		`,
 	})
 	validate(t, &testCase{
@@ -191,19 +194,24 @@ func TestFormatStringCSV(t *testing.T) {
 			"ModelA": Assessments{
 				AssessmentKeyCoverageStatement: 1,
 				AssessmentKeyFilesExecuted:     2,
-				AssessmentKeyResponseNoExcess:  4,
+				AssessmentKeyResponseNoError:   4,
+				AssessmentKeyResponseNoExcess:  5,
+				AssessmentKeyResponseNotEmpty:  6,
+				AssessmentKeyResponseWithCode:  7,
 			},
 			"ModelB": Assessments{
 				AssessmentKeyCoverageStatement: 1,
-				AssessmentKeyFilesExecuted:     2,
-				AssessmentKeyResponseNoExcess:  4,
+				AssessmentKeyFilesExecuted:     2, AssessmentKeyResponseNoError: 4,
+				AssessmentKeyResponseNoExcess: 5,
+				AssessmentKeyResponseNotEmpty: 6,
+				AssessmentKeyResponseWithCode: 7,
 			},
 		},
 
 		ExpectedString: `
-			model,coverage-statement,files-executed,response-no-excess
-			ModelA,1,2,4
-			ModelB,1,2,4
+			model,coverage-statement,files-executed,response-no-error,response-no-excess,response-not-empty,response-with-code
+			ModelA,1,2,4,5,6,7
+			ModelB,1,2,4,5,6,7
 		`,
 	})
 }
--- a/evaluate/metrics/assessment_test.go
+++ b/evaluate/metrics/assessment_test.go
@@ -39,11 +39,11 @@ func TestAssessmentsAdd(t *testing.T) {
 
 		Assessments: NewAssessments(),
 		X: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
 		},
 
 		ExpectedAssessments: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
 		},
 	})
 
@@ -51,19 +51,19 @@ func TestAssessmentsAdd(t *testing.T) {
 		Name: "Existing key",
 
 		Assessments: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
 		},
 		X: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
 		},
 
 		ExpectedAssessments: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 2,
+			AssessmentKeyResponseNoExcess: 2,
 		},
 	})
 }
 
-func TestMerge(t *testing.T) {
+func TestAssessmentsMerge(t *testing.T) {
 	type testCase struct {
 		Name string
 
@@ -92,11 +92,11 @@ func TestMerge(t *testing.T) {
 
 		A: NewAssessments(),
 		B: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
 		},
 
 		ExpectedC: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
 		},
 	})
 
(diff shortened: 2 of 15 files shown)


model/llm/prompt/parse.go
```diff
@@ -17,28 +17,39 @@ var (
...
+
 	// Some models produce duplicated code tags, so unify them if needed.
 	response = codeTagDuplicatedMatch.ReplaceAllString(response, "```")
 
 	blocks := bytesutil.GuardedBlocks(response, codeTagMatch, codeTagMatch)
 
 	// When no code blocks are found, assume that just the code is returned.
 	if len(blocks) == 0 {
-		assessment[metrics.AssessmentKeyNoExcessResponse] = 1
+		// If we cannot distinguish between code and text, we sadly also cannot check if the response contains actual code or if there is any excess response content.
```
bauersimon, line 32: In the future we should extend the language interface to maybe contain a function that checks if some string resembles syntactically valid code.

zimmski: I think that is a super hard problem.

evaluate/metrics/assessment_test.go
```diff
@@ -39,31 +39,31 @@ func TestAssessmentsAdd(t *testing.T) {
 
 		Assessments: NewAssessments(),
 		X: map[AssessmentKey]uint{
-			AssessmentKeyNoExcessResponse: 1,
+			AssessmentKeyResponseNoExcess: 1,
```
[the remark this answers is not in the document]
bauersimon: Right. Wanted to do that... thought maybe with adding the other assessments it's clear to see why we are renaming anyways.

evaluate/metrics/assessment.go
```diff
@@ -57,6 +63,24 @@ func (a Assessments) Add(x Assessments) {
 	}
 }
 
+// IsEqual checks if both assessment collections are equal.
+func (a Assessments) IsEqual(x Assessments) bool {
+	if a == nil {
```
zimmski, line 66: Better not allocate, its wasteful. Add a 

if len(x) != len(y) { return false }

at the top. handles the nil cases right away.

evaluate/metrics/assessment.go
```diff
@@ -57,6 +63,24 @@ func (a Assessments) Add(x Assessments) {
 	}
 }
 
+// IsEqual checks if both assessment collections are equal.
+func (a Assessments) IsEqual(x Assessments) bool {
```
zimmski, line 65: Tests

zimmski: Nope. they are there, didn't see them.

evaluate/metrics/assessment.go
zimmski: IsEqual -> Equal

Like in assert.Equal.

evaluate/metrics/assessment_test.go
```diff
@@ -134,24 +134,27 @@ func TestAssessmentString(t *testing.T) {
...
 
 		Assessment: NewAssessments(),
 
-		ExpectedString: "files-executed=0, files-problems=0, coverage-statement=0, response-no-excess=0",
+		ExpectedString: "files-executed=0, files-problems=0, coverage-statement=0, response-no-error=0, response-not-empty=0, response-with-code=0, response-no-excess=0",
 	})
 
 	validate(t, &testCase{
-		Name: "Empty Metrics",
+		Name: "Value Metrics",
```
zimmski, line 145: Why not "Non-empty metrics"?

evaluate/metrics/assessment_test.go
```diff
@@ -134,24 +134,27 @@ func TestAssessmentString(t *testing.T) {
...
 			AssessmentKeyFilesProblems:     3,
-			AssessmentKeyResponseNoExcess:  4,
+			AssessmentKeyResponseNoError:   4,
+			AssessmentKeyResponseNoExcess:  5,
+			AssessmentKeyResponseNotEmpty:  6,
+			AssessmentKeyResponseWithCode:  7,
 		},
 
-		ExpectedString: "files-executed=2, files-problems=3, coverage-statement=1, response-no-excess=4",
+		ExpectedString: "files-executed=2, files-problems=3, coverage-statement=1, response-no-error=4, response-not-empty=6, response-with-code=7, response-no-excess=5",
```
zimmski: Wondering. Headers right now are not deterministic. let's sort them by alphabet right away?

evaluate/metrics/assessment_test.go
```diff
@@ -193,20 +196,103 @@ func TestFormatStringCSV(t *testing.T) {
...
-			ModelA,2,3,1,4
-			ModelB,2,3,1,4
+			model,files-executed,files-problems,coverage-statement,response-no-error,response-not-empty,response-with-code,response-no-excess
+			ModelA,2,3,1,4,6,7,5
+			ModelB,2,3,1,4,6,7,5
 		`,
 	})
 }
+
+func TestAssessmentsIsEqual(t *testing.T) {
```
zimmski, line 220: https://github.com/symflower/eval-dev-quality/pull/39/commits/a1d5ee333f333310d1af796e6a6c727c12bf8954#r1570186697

evaluate/metrics/assessment_test.go
```diff
@@ -193,20 +196,103 @@ func TestFormatStringCSV(t *testing.T) {
...
+	type testCase struct {
+		Name string
+
+		Assessments Assessments
+		X           Assessments
+
+		ExpectedBool bool
```
zimmski, line 227: I see you used Symflower for that one as well 🧐

evaluate/metrics/assessment_test.go
bauersimon: They are. The ordering is always the same as how the assessment keys are registered.

evaluate/metrics/assessment_test.go
```diff
@@ -193,20 +196,103 @@ func TestFormatStringCSV(t *testing.T) {
...
+		},
+		X: Assessments{
+			AssessmentKeyResponseWithCode: 2,
+		},
+
+		ExpectedBool: true,
+	})
+
+	validate(t, &testCase{
+		Name: "Default Value",
```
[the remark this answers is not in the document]
bauersimon: yeah because the default value for any key that is not set is zero :thinking:

evaluate/metrics/assessment_test.go
bauersimon: linking to comments does not work at all for me... it always sends me to the whole changes... I think the link might only work once the review is sent?

evaluate/metrics/assessment_test.go
zimmski: Not if you are registering keys from other packages. That is IMO where we are heading: common task-packages that define their keys.

evaluate/metrics/assessment_test.go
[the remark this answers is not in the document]
zimmski: i am unsure, in one set the key exists and in the other it doesn't. maybe that is equal for metrics, maybe not. Unsure... Tag Andi for third opinion?

evaluate/metrics/assessment_test.go
zimmski: https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570186697

zimmski: Looks like it. but i always send the comment right away. i do not use the start-review-functionality. because i hate that even more.

evaluate/repository.go
```diff
@@ -75,6 +76,11 @@ func EvaluateRepository(resultPath string, model model.Model, language language.
 		if coverage == 100 {
 			repositoryAssessment[metrics.AssessmentKeyCoverageStatement]++
 		}
+
+		// If we weren't able to determine if the response contained any code but the execution was successful, correct that now by incrementing the "AssessmentKeyResponseWithCode".
+		if generationAssessments[metrics.AssessmentKeyResponseWithCode] == 0 {
+			repositoryAssessment[metrics.AssessmentKeyResponseWithCode]++
```
zimmski, line 80: Would not do that here. Should be already determined by GenerateTestsForFile

evaluate/repository.go
```diff
@@ -55,14 +55,15 @@ func EvaluateRepository(resultPath string, model model.Model, language language.
 
 	repositoryAssessment = metrics.NewAssessments()
 	for _, filePath := range filePaths {
-		assessments, err := model.GenerateTestsForFile(language, temporaryRepositoryPath, filePath)
+		generationAssessments, err := model.GenerateTestsForFile(language, temporaryRepositoryPath, filePath)
```
zimmski, line 58: Please refactor that loop-body into a function and write a test for it.

model/llm/llm_test.go
```diff
@@ -46,7 +46,7 @@ func TestModelLLMGenerateTestsForFile(t *testing.T) {
 
 			actualAssessment, actualError := llm.GenerateTestsForFile(tc.Language, temporaryPath, tc.SourceFilePath)
 			assert.NoError(t, actualError)
-			assert.Equal(t, tc.ExpectedAssessment, actualAssessment)
+			assert.Truef(t, tc.ExpectedAssessment.IsEqual(actualAssessment), "expected:%s\nactual:%s", tc.ExpectedAssessment, actualAssessment)
```
zimmski, line 49: Why are we changing this assert?

zimmski: Looks like reimplementing ".Equal"

model/llm/llm_test.go
```diff
@@ -89,6 +89,8 @@ func TestModelLLMGenerateTestsForFile(t *testing.T) {
 
 		ExpectedAssessment: metrics.Assessments{
 			metrics.AssessmentKeyResponseNoExcess: 1,
+			metrics.AssessmentKeyResponseNotEmpty: 1,
+			metrics.AssessmentKeyResponseWithCode: 1,
```
zimmski, line 93: I wish we had a "automatic sort" tool in Symflower

model/llm/prompt/parse.go
```diff
@@ -17,28 +17,39 @@ var (
...
+	if isOnlyWhitespace(responseWithoutBlock) {
 		assessment[metrics.AssessmentKeyResponseNoExcess] = 1
 	} else {
 		assessment[metrics.AssessmentKeyResponseNoExcess] = 0
 	}
 
 	return assessment, strings.TrimSpace(codeTagMatch.ReplaceAllString(block, ""))
 }
+
+// isOnlyWhitespace checks if the string contains only whitespace as defined by unicode.
```
zimmski, line 52: unicode -> Unicode

Common name.

model/llm/prompt/parse_test.go
```diff
@@ -23,7 +23,7 @@ func TestParseResponse(t *testing.T) {
 		t.Run(tc.Name, func(t *testing.T) {
 			actualAssessment, actualCode := ParseResponse(tc.Response)
 
-			assert.Equal(t, tc.ExpectedAssessment, actualAssessment)
+			assert.Truef(t, tc.ExpectedAssessment.IsEqual(actualAssessment), "expected:%s\nactual:%s", tc.ExpectedAssessment, actualAssessment)
```
zimmski, line 26: Same as https://github.com/symflower/eval-dev-quality/pull/39/commits/a1d5ee333f333310d1af796e6a6c727c12bf8954#r1570229136

model/symflower/symflower_test.go
```diff
@@ -49,7 +49,7 @@ func TestModelSymflowerGenerateTestsForFile(t *testing.T) {
 			} else if actualError != nil || tc.ExpectedErrorText != "" {
 				assert.ErrorContains(t, actualError, tc.ExpectedErrorText)
 			}
-			assert.Equal(t, tc.ExpectedAssessment, actualAssessment)
+			assert.Truef(t, tc.ExpectedAssessment.IsEqual(actualAssessment), "expected:%s\nactual:%s", tc.ExpectedAssessment, actualAssessment)
```
zimmski, line 52: https://github.com/symflower/eval-dev-quality/pull/39/commits/a1d5ee333f333310d1af796e6a6c727c12bf8954#r1570229136

model/llm/llm_test.go
bauersimon: because `.Equal` also checks if keys exist but that should not matter for our assessments IMO, see https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570210755

model/llm/prompt/parse.go
```diff
@@ -17,28 +17,39 @@ var (
 func ParseResponse(response string) (assessment metrics.Assessments, code string) {
 	assessment = metrics.Assessments{}
 
+	if isOnlyWhitespace(response) {
+		return assessment, response
+	}
+	assessment[metrics.AssessmentKeyResponseNotEmpty] = 1
```
zimmski, line 23: Please always use "+=" for all metrics. If you set it to "0" same thing `+= 0` we are adding to the assessment. If there are multiple additions, we then don't have to handle setting them, because they are just added.

model/llm/prompt/parse.go
```diff
@@ -17,28 +17,39 @@ var (
...
 	response = codeTagDuplicatedMatch.ReplaceAllString(response, "```")
 
 	blocks := bytesutil.GuardedBlocks(response, codeTagMatch, codeTagMatch)
 
 	// When no code blocks are found, assume that just the code is returned.
 	if len(blocks) == 0 {
-		assessment[metrics.AssessmentKeyNoExcessResponse] = 1
+		// If we cannot distinguish between code and text, we sadly also cannot check if the response contains actual code or if there is any excess response content.
 
 		return assessment, strings.TrimSpace(response)
```
zimmski, line 34: Why are we trimming the response here again? Let's just do that once above where we do the empty check.

model/llm/prompt/parse_test.go
```diff
@@ -38,13 +38,26 @@ func TestParseResponse(t *testing.T) {
...
+
 	validate(t, &testCase{
 		Name: "Only Code",
 
 		Response: code,
 
 		ExpectedAssessment: metrics.Assessments{
-			metrics.AssessmentKeyResponseNoExcess: 1,
+			metrics.AssessmentKeyResponseNotEmpty: 1,
+			metrics.AssessmentKeyResponseNoExcess: 0,
```
zimmski, line 59: Since it is only code, shouldn't `AssessmentKeyResponseNoExcess:` be set to 1 then?

model/llm/prompt/parse_test.go
```diff
@@ -38,13 +38,26 @@ func TestParseResponse(t *testing.T) {
...
+			metrics.AssessmentKeyResponseWithCode: 0,
```
zimmski, line 60: Since it is only code, shouldn't this be set to 1?

provider/openrouter/openrouter.go
```diff
@@ -56,8 +55,8 @@ func (p *openRouterProvider) Query(ctx context.Context, modelIdentifier string,
 	)
 	if err != nil {
 		return "", pkgerrors.WithStack(err)
-	} else if len(apiResponse.Choices) == 0 {
-		return "", fmt.Errorf("empty LLM %q response: %+v", modelIdentifier, apiResponse)
+	} else if len(apiResponse.Choices) == 0 { // Convert no choices to an empty response.
```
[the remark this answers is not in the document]
bauersimon: Should we then check for this error and introduce an extra assessment? Or just treat it as an error.

model/llm/prompt/parse_test.go
```diff
@@ -67,7 +82,9 @@ func TestParseResponse(t *testing.T) {
 			Response: "Some text...\n\n```\n" + code + "\n```\n\nSome more text...",
 
 			ExpectedAssessment: metrics.Assessments{
+				metrics.AssessmentKeyResponseNotEmpty: 1,
 				metrics.AssessmentKeyResponseNoExcess: 0,
```
zimmski, line 86: SInce there is "Some more text", excess should be 1?

provider/openrouter/openrouter.go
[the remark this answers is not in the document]
zimmski: It is an error of the model. So the error metric should increment. So we kind of need a ResponseNoError or something?

model/llm/llm_test.go
zimmski: I see. can you introduce .Equal then with its own commit and add a helper to make that clear with a commit why we need it.

model/llm/prompt/parse_test.go
bauersimon: Yes. But it does not have a code fence. So we cannot decide if its code or not.

That's what the comment `// If we cannot distinguish between code and text, we sadly also cannot check if the response contains actual code or if there is any excess response content.` says.

Then we are back at this https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1568814274 problem.

I'm trying to correct for such cases with https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570227179.

evaluate/repository.go
bauersimon: We don't determine some cases (at least for now) within `GenerateTestsForFile` right now.

See https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570282777.

We can remove this and accept the fact that we might have some incorrect scoring for now.

model/llm/prompt/parse_test.go
zimmski: I see. Can you comment the test case as well please.

The correction of https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570227179 does not make sense to me. Either the function that analyses the response says it is code or not. Should not be determined afterwards. Setting that metric later, is a behavioral bug to me.

model/llm/prompt/parse.go
bauersimon: Then I cannot have a helper function for the empty check :thinking: but fine with me as well.

model/llm/prompt/parse_test.go
bauersimon: Same as https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570282777

evaluate/metrics/assessment_test.go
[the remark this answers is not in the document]
bauersimon: Then we rename the function. But for scoring purposes these should be considered equal IMO. @ahumenberger what do you think?

provider/openrouter/openrouter.go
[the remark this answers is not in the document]
bauersimon: that already exists... so this just stays an error then as before

model/llm/prompt/parse_test.go
bauersimon: Alrightsy. Will comment and remove the hacky correction afterwards. Until we have some way of detecting if it the response resembles source code. We will have scoring issues then (where the execution runs fine but we don't detect that it's source code).

evaluate/metrics/assessment.go
bauersimon: > if len(x) != len(y) { return false }

But that one is incorrect if one assessment explicitly sets a key to `0` and the other one does not set it, meaning we have the default value `0` so they end up being equal. (https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570206152)

bauersimon: Changed it to:

```go
if a == nil || x == nil {
	return a == nil && x == nil
}
```

model/llm/prompt/parse_test.go
zimmski: Does that mean, we cannot replicate the manual-log-checking-results? Or what is the problem? Don't get it.

evaluate/metrics/assessment.go
zimmski: Ah i see

model/llm/prompt/parse_test.go
bauersimon: No - there is additional text. So no points for `NoExcess`. It's correct the way it is.

model/llm/prompt/parse_test.go
bauersimon: The problem is when the model does **not** respond with code fences. We currently have no way of telling within `GenerateTestsForFile` if the response is code or just text (https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1568814274).

In the manual log checking we were able to decide if a model was trying to produce code or just text easily by just looking at it.

Because of goodwill, we currently assume that a response without code fences is code and forward it to the execution (cause that will of course fail if it's not code). I was trying to correct for these errors in hindsight in a hacky way: https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1570227179 (basically awarding points in hindsight after the execution ran successfully).

model/llm/prompt/parse_test.go
zimmski: AH, right! Negative!

model/llm/prompt/parse.go
```diff
@@ -17,28 +17,39 @@ var (
...
 		assessment[metrics.AssessmentKeyResponseNoExcess] = 1
 	} else {
 		assessment[metrics.AssessmentKeyResponseNoExcess] = 0
 	}
 
 	return assessment, strings.TrimSpace(codeTagMatch.ReplaceAllString(block, ""))
 }
+
+// isOnlyWhitespace checks if the string contains only whitespace as defined by unicode.
+func isOnlyWhitespace(data string) bool {
```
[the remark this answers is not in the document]
bauersimon: We already have that in `bytesutil` ._.

model/llm/prompt/parse_test.go
zimmski: > So to be as good as manual categorization. We must implement [#39 (comment)](https://github.com/symflower/eval-dev-quality/pull/39#discussion_r1568814274) right away.

Is that the only problem we have? Then let's just document the heck out of it in the implementation+test and move on. If we can do every result automatically except for those few, we can check what that means and if there are too many lost, then we iterate.

evaluate/repository.go
bauersimon: Follow-up?

evaluate/metrics/assessment_test.go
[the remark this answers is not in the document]
ahumenberger: I think for metrics it's fine. As long as "being in the set" has no special meaning, there should be no problem.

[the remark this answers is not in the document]
ahumenberger: Another way to think about it: If it's possible to turn the map into a struct with fields, then there is no difference between not being in the set, and being in the set with zero.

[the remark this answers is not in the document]
zimmski: Makes sense. With that view, it is not a set. Which is better to think about for metrics.

util/slices_test.go
```diff
@@ -0,0 +1,71 @@
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInsertToSortedSlice(t *testing.T) {
```
zimmski, line 9: Noice!

zimmski (approved):

zimmski:
Super nice!

PR #40, opened 2024-04-18
Test "evaluate" package

zimmski:


evaluate/repository_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/symflower/eval-dev-quality/evaluate/metrics"
+	"github.com/symflower/eval-dev-quality/language"
+	"github.com/symflower/eval-dev-quality/model"
+	"github.com/symflower/eval-dev-quality/model/symflower"
+	"github.com/zimmski/osutil"
```
bauersimon, line 13: split imports

evaluate/repository_test.go
```diff
@@ -0,0 +1,68 @@
...
+		TestDataPath:   "../testdata",
+		RepositoryPath: "golang/plain",
+
+		ExpectedRepositoryAssessment: metrics.Assessments{
+			metrics.AssessmentKeyCoverageStatement: 1,
+			metrics.AssessmentKeyFilesExecuted:     1,
+			metrics.AssessmentKeyNoExcessResponse:  1,
+		},
+		ExpectedResultFiles: []string{
+			"symflower_symbolic-execution/golang/golang/plain.log",
```
bauersimon, line 65: huh so we have golang once from the language and once because its in the repository path :thinking:

bauersimon (changes requested):

evaluate/repository_test.go
zimmski: Yarg. Time to get the Symflower-linter into the symflower binary so we can use it here.

evaluate/repository_test.go
zimmski: Yes. I think it is fine that way for now.

zimmski: Reason is that we are not yet set in stone with how we manage the repositories. Could be that we merge them because it is then easier.

bauersimon (approved):

PR #46, opened 2024-04-18
Integration testing for the binary

zimmski:


model/symflower/symflower_test.go
```diff
@@ -33,6 +33,13 @@ func TestModelSymflowerGenerateTestsForFile(t *testing.T) {
 
 	validate := func(t *testing.T, tc *testCase) {
 		t.Run(tc.Name, func(t *testing.T) {
+			log, logger := log.Buffer()
+			defer func() {
+				if t.Failed() {
+					assert.Empty(t, log.String())
```
bauersimon, line 39: Could probably also do `t.Log(log.String())` to avoid an extra reported failed assertion.

zimmski: That is the point. If it fails, give extra data

bauersimon: No yeah... but it will report the log as "expected empty" which might be confusing in the future. Just using `t.Log` will also print the log but without an "artificial" error message.

zimmski: ah!

zimmski: Sorry misunderstood! Done, Please have another look!

bauersimon (approved):

PR #48, opened 2024-04-18
Categorization

bauersimon:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -109,12 +109,14 @@ func (command *Evaluate) Execute(args []string) (err error) {
 
 	// Check that models and languages can be evaluated by executing the "plain" repositories.
 	log.Printf("Checking that models and languages can be used for evaluation")
-	assessmentsPerModel := map[string]metrics.Assessments{}
+	assessmentsPerModelPerRepository := map[string]map[string]metrics.Assessments{}
```
zimmski, line 112: Should we split by "language" as well right away?

zimmski, line 112: Should we introduce identifier types for models, languages, ... ? Would make them easier to read and allows to not mix? Kind ... of ... type ... safety?

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -163,7 +165,7 @@ func (command *Evaluate) Execute(args []string) (err error) {
 				language := language.Languages[languageID]
 
 				assessment, ps, err := evaluate.EvaluateRepository(command.ResultPath, model, language, command.TestdataPath, filepath.Join(languageID, repository.Name()))
-				assessmentsPerModel[model.ID()].Add(assessment)
+				assessmentsPerModelPerRepository[model.ID()][repository.Name()] = metrics.Merge(assessmentsPerModelPerRepository[model.ID()][repository.Name()], assessment)
```
zimmski, line 168: Please use and .Add

zimmski: The bigger these maps get, the worst is the copying

zimmski: And it is then easier to read what we do here

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -172,13 +174,21 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		}
 	}
 
+	// Combine multiple repository assessments for reporting.
+	assessmentsPerModel := make(map[string]metrics.Assessments, len(assessmentsPerModelPerRepository))
```
zimmski, line 178: Can you print assessmentsPerModelPerRepository before to the CSV and then do the logic for assessmentsPerModel in one block to print the scores.

zimmski: I think that is better grouping: print overall results first and then combine into different reports.

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: The assessments per repository are never initialized so the object to `Add` to is `nil`. So I guess I need to manually initialize all of them first?

zimmski: We are doing that for "plain" as well, so i say, yes?

evaluate/metrics/assessment.go
```diff
@@ -137,28 +137,33 @@ func (a Assessments) StringCSV() (row []string) {
 }
 
 func csvHeader() []string {
```
zimmski, line 139: document

bauersimon:
todo: add category unknown if total score is not known

evaluate/metrics/assessment.go
```diff
@@ -137,28 +137,33 @@ func (a Assessments) StringCSV() (row []string) {
...
-	return append([]string{"model", "score"}, allAssessmentKeysStrings...)
+	return append([]string{"model", "repository", "score"}, allAssessmentKeysStrings...)
 }
 
 // FormatStringCSV formats the given assessment metrics as CSV.
```
zimmski, line 143: Should we call this just `FormatCSV` it is the only one. so no need to give it a type

evaluate/metrics/assessment_test.go
```diff
@@ -179,41 +179,47 @@ func TestFormatStringCSV(t *testing.T) {
 	validate(t, &testCase{
 		Name: "Single Empty Model",
 
-		AssessmentPerModel: map[string]Assessments{
-			"Model": Assessments{},
+		AssessmentPerModelPerRepository: map[string]map[string]Assessments{
+			"Model": map[string]Assessments{
+				"Repository": NewAssessments(),
```
zimmski, line 184: usually when we do not care we always prefix with "some" -> `some-repository`

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -182,8 +182,12 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		}
 	}
 
+	isPlain := len(commandRepositories) == 0 || (len(commandRepositories) == 1 && commandRepositories["plain"])
```
zimmski, line 185: isPlain -> isPlainRepository

zimmski: Wait. what is the logic here? Why do we not ccategorize every repository? I know that we currently only have "plain" Go repository, but why do we already block new ones here?

zimmski: (don't think that this is necessary)

evaluate/metrics/category.go
```diff
@@ -0,0 +1,42 @@
...
+	// AssessmentCategoryResponseNoCode indicates that a model's response did not contain any source code.
+	AssessmentCategoryResponseNoCode = AssessmentCategory("response-no-code")
+	// AssessmentCategoryCodeInvalid indicates that a model's generated code produced an error when executed.
+	AssessmentCategoryCodeInvalid = AssessmentCategory("code-invalid")
+	// AssessmentCategoryCodeExecuted indicates that a model's generated code could be executed without an error.
+	AssessmentCategoryCodeExecuted = AssessmentCategory("code-executed")
+	// AssessmentCategoryCodeCoverageStatementReached indicates that a model's generated code reached 100% statement coverage.
+	AssessmentCategoryCodeCoverageStatementReached = AssessmentCategory("code-coverage-statement")
+	// AssessmentCategoryCodeNoExcess indicates that a model's response did not contain more content than requested.
+	AssessmentCategoryCodeNoExcess = AssessmentCategory("code-no-excess")
```
zimmski, line 20: Perfect category is missing?

zimmski: I guess it is `AssessmentCategoryCodeNoExcess`?

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: cause for others we don't know the total score... will introduce a category unknown in that case

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: Could also use the models and languages directly instead of IDs. That way its type safe already.

zimmski: i mean more in the sense of something like this

someMap[modelID]...

would then fail if the key of the map is of type repositoryID

bauersimon: Could also work. But why introduce extra types if we could just do:

```go
// AssessmentPerModelPerLanguagePerRepository holds a collection of assessments per model per language and per repository.
type AssessmentPerModelPerLanguagePerRepository map[model.Model]map[language.Language]map[string]Assessments
```

zimmski: Sure!

language/testing/mock.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"log"
+
+	"github.com/stretchr/testify/mock"
+
+	"github.com/symflower/eval-dev-quality/language"
+)
+
+// MockLanguage is a mocked language.
+type MockLanguage struct {
```
zimmski, line 12: Holy wow. that was a big commit!

zimmski (approved):

PR #55, opened 2024-04-19
refactor, Consistent naming conventions for languages, models and providers

zimmski:


model/symflower/symflower_test.go
```diff
@@ -11,14 +11,15 @@ import (
...
+	"github.com/symflower/eval-dev-quality/language/golang"
 	"github.com/symflower/eval-dev-quality/log"
 )
 
 func TestModelSymflowerGenerateTestsForFile(t *testing.T) {
 	type testCase struct {
 		Name string
 
-		ModelSymflower *ModelSymflower
+		ModelSymflower *Model
```
bauersimon, line 22: maybe also rename the field?

language/golang/golang_test.go
```diff
@@ -10,14 +10,15 @@ import (
...
 	"github.com/symflower/eval-dev-quality/log"
 )
 
-func TestLanguageGolangFiles(t *testing.T) {
+func TestLanguageFiles(t *testing.T) {
 	type testCase struct {
 		Name string
 
-		LanguageGolang *LanguageGolang
+		LanguageGolang *Language
```
bauersimon, line 21: maybe rename the field

language/golang/golang_test.go
```diff
@@ -47,19 +48,19 @@ func TestLanguageGolangFiles(t *testing.T) {
...
 	})
 }
 
-func TestLanguageGolangExecute(t *testing.T) {
+func TestLanguageExecute(t *testing.T) {
 	type testCase struct {
 		Name string
 
-		LanguageGolang *LanguageGolang
+		LanguageGolang *Language
```
bauersimon, line 63: maybe rename the field

model/symflower/symflower.go
```diff
@@ -12,18 +12,23 @@ import (
...
+// Model holds a Symflower model using the locally installed CLI.
+type Model struct{}
 
-var _ model.Model = (*ModelSymflower)(nil)
+// NewModel returns a Symflower model.
+func NewModel() (model model.Model) {
+	return &Model{}
+}
+
+var _ model.Model = (*Model)(nil)
```
bauersimon, line 23: no real need to keep that check here since the function above contains the same type scenario (trying to return a model instance as a `model.Model`.

similar for a lot of other cases in this commit

zimmski: Its a pattern for me that we do internally as well: explicitly state the interface you want to implement. then put the methods afterwards.

bauersimon (approved):

PR #56, opened 2024-04-19
fix, Lock Symflower installations to make sure two evalatuions do not install at the same time

zimmski:


bauersimon (commented):
trusting that this works \o/

bauersimon (approved):

PR #60, opened 2024-04-22
Let language decide test file paths instead of hardcoding that

### LINKED ISSUE #58: Don't hardcode test file path
For LLM models we currently hardcode the test file path. This does not scale.

- [x] let language decide what the test file path is

bauersimon:
Closes #58

### DIFF
--- a/language/golang/golang.go
+++ b/language/golang/golang.go
@@ -59,6 +59,12 @@ func (l *Language) Files(log *log.Logger, repositoryPath string) (filePaths []st
 	return filePaths, nil
 }
 
+// TestFilePath returns the file path of a test file given the corresponding file path of the test's source file.
+func (l *Language) TestFilePath(projectRootPath string, filePath string) (testFilePath string) {
+	return strings.TrimSuffix(filePath, ".go") + "_test.go"
+
+}
+
 var languageGoNoTestsMatch = regexp.MustCompile(`(?m)^DONE (\d+) tests.*in (.+?)$`)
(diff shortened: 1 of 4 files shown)


language/golang/golang.go
```diff
@@ -59,6 +59,13 @@ func (l *Language) Files(log *log.Logger, repositoryPath string) (filePaths []st
 	return filePaths, nil
 }
 
+// TestFilePath returns the file path of a test file given the corresponding file path of the test's source file.
+// The file paths may be relative.
+func (l *Language) TestFilePath(filePath string) (testFilePath string) {
```
zimmski, line 64: Please add the repository path as well. The file path of Java projects is dependent on the projects path.

language/golang/golang.go
```diff
@@ -59,6 +59,13 @@ func (l *Language) Files(log *log.Logger, repositoryPath string) (filePaths []st
 	return filePaths, nil
 }
 
+// TestFilePath returns the file path of a test file given the corresponding file path of the test's source file.
```
zimmski, line 62: Sounds complicated, why not?

-> `TestFilePath returns the test file path of a implementation file path.`

language/golang/golang.go
bauersimon: Mhm or we change the docstring that the filepath always needs to be relative to the project root?

bauersimon: but just adding the project root is probably fine as well

zimmski: All our parameters are always relative to their context. So would not mention that in the doc

zimmski (approved):

PR #62, opened 2024-04-22
Java language implementation

zimmski:


language/java/language_test.go
```diff
@@ -0,0 +1,220 @@
...
+		})
+	}
+
+	validate(t, &testCase{
+		Name: "Source file",
+
+		FilePath: "src/main/java/com/eval/Plain.java",
+
+		ExpectedTestFilePath: "src/test/java/com/eval/PlainTest.java",
+	})
```
bauersimon, line 122: Would be nice to see also a case where the `src/main/java` is not there. And also one with a root path.

language/java/language_test.go
```diff
@@ -0,0 +1,220 @@
...
+		})
+	}
+
+	validate(t, &testCase{
+		Name: "Source file",
+
+		FilePath: "src/main/java/com/eval/Plain.java",
+
+		ExpectedImportPath: "com.eval",
+	})
```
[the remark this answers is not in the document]
zimmski: With moving fully to "symflower" commands, this will be handled automatically. Just want to have the case implemented for the plain.java now.

testdata/java/plain/pom.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- REMARK Do not edit this file as it was generated by `symflower-gen language-testdata-repositories`. -->
```
bauersimon, line 2: Guess copy paste error?

language/java/language.go
```diff
@@ -0,0 +1,117 @@
...
+		t := "src/test/java"
+		filePath = filePath[:l] + t + filePath[l+len(t):]
+	}
+
+	return strings.TrimSuffix(filePath, ".java") + "Test.java"
+}
+
+// TestFramework returns the human-readable name of the test framework that should be used.
+func (l *Language) TestFramework() (testFramework string) {
+	return "JUnit5"
```
bauersimon, line 87: Shouldn't this be "JUnit 5" now, then?

language/java/language.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return importPath
+}
+
+// TestFilePath returns the file path of a test file given the corresponding file path of the test's source file.
+func (l *Language) TestFilePath(projectRootPath string, filePath string) (testFilePath string) {
+	if l := strings.LastIndex(filePath, "src/main/java"); l != -1 {
+		t := "src/test/java"
+		filePath = filePath[:l] + t + filePath[l+len(t):]
+	}
```
bauersimon, line 80: Just noticed that it's probably simpler to do `strings.Replace(filePath, "src/main/java", "src/test/java")` here...

bauersimon (changes requested):

language/java/language.go
zimmski: Its fine. We should use Symflowers functionality anyway with the next version.

evaluate/report/collection.go
```diff
@@ -27,6 +28,10 @@ func NewAssessmentPerModelPerLanguagePerRepository(models []model.Model, languag
 				a[m][l] = map[string]metrics.Assessments{}
 			}
 			for _, r := range repositories {
+				if !strings.HasPrefix(r, l.ID()) { // Ensure the repository path matches the language.
```
zimmski, line 31: Would always do a block comment for something like this, unless you are directly commenting the condition.

language/golang/language.go
```diff
@@ -82,14 +83,9 @@ var languageGoNoCoverageMatch = regexp.MustCompile(`(?m)^coverage: \[no statemen
...
 		Command: []string{
-			"gotestsum",
-			"--format", "standard-verbose", // Keep formatting consistent.
-			"--hide-summary", "skipped", // We are not interested in skipped tests, because they are the same as no tests at all.
-			"--",       // Let the real Go "test" tool options begin.
-			"-cover",   // Enable statement coverage.
-			"-v",       // Output with the maximum information for easier debugging.
-			"-vet=off", // Disable all linter checks, because those should be part of a different task.
-			"./...",    // Always execute all tests of the repository in case multiple test files have been generated.
+			tools.SymflowerPath, "test",
```
zimmski, line 86: Using the newer Symflower version needs to be done, but we can only do that, when we have it published.

zimmski: This is finally merged and on its way to be published...

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -97,11 +98,28 @@ func (command *Evaluate) Execute(args []string) (err error) {
 	}
 
 	commandRepositories := map[string]bool{}
+	commandRepositoriesLanguages := map[string]bool{}
 	for _, r := range command.Repositories {
-		commandRepositories[r] = true
+		ds := strings.SplitN(r, string(os.PathSeparator), 2)
+		languageIDOfRepository := ds[0]
```
bauersimon, line 104: nit

could inline because I'm not even sure what `ds` stands for...

PR #66, opened 2024-04-23
Generate mocks so there is no more manual work when our interfaces change

bauersimon:


Makefile
```diff
@@ -39,10 +39,15 @@ install-all: install install-tools-testing # Install everything for and of this
...
 	go install -v gotest.tools/gotestsum@v1.11.0
 	eval-dev-quality install-tools
 .PHONY: install-tools-testing
 
-test: # [<Go package] - # Test everything, or only the specified package.
+generate: # Run code generation.
+	mockery
+.PHONY: generate
+
+test: generate # [<Go package] - # Test everything, or only the specified package.
```
zimmski, line 51: Would not generate on every test run but only if necessary.

zimmski, line 51: Please add a CI check that fails if we the generation is not up to date. Y ou can copy that from our internal lint-build CI job.

zimmski (approved):

PR #69, opened 2024-04-24
Early merger for #62

zimmski:


zimmski:
@bauersimon we already reviewed these commits and they can be merged right away so making progress.

PR #70, opened 2024-04-25
fix, Allow whitespace after code tags to properly detect code blocks, and split LLM logging into request and response

bauersimon:


language/testing/Language_mock_gen.go
```diff
@@ -3,9 +3,8 @@
 package languagetesting
 
 import (
-	log "log"
-
 	mock "github.com/stretchr/testify/mock"
+	log "github.com/symflower/eval-dev-quality/log"
```
zimmski, line 7: Just wondering: how are we defining the log package? Because when you do "make generate", we overwrite it again?

zimmski: No. tried it.

log/logger.go
```diff
@@ -11,12 +11,21 @@ import (
 )
 
 // Logger holds a logger to log to.
-type Logger = log.Logger
+type Logger struct {
```
zimmski, line 14: Please revert that. We want to alias that type. not embed it. `type Logger = log.Logger` is fine.

bauersimon:
lgtm

PR #89, opened 2024-05-02
Award 10 points for reaching maximal coverage instead of 1 and note result differences in evaluation README

bauersimon:


zimmski:
10 points to Gryffindor for a PR well done

zimmski:
Putting this in the v0.4.0 milestone, because we used it there.

PR #94, opened 2024-05-06
Additional CSVs to sum up metrics for all models overall and per language

Munsio:
@bauersimon I am not sure if we want to do early mergers or not but I think you can still review this one already.

I've used 2 refactoring commits for the changes to `AssessmentPerModel` for easier reviewing.

### DIFF
--- a/cmd/eval-dev-quality/cmd/evaluate.go
+++ b/cmd/eval-dev-quality/cmd/evaluate.go
@@ -274,7 +274,7 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		totalScore = uint(len(languagesSelected))
 	}
 
-	assessmentsPerModel := assessments.Collapse()
+	assessmentsPerModel := assessments.CollapseByModel()
 	if err := (report.Markdown{
 		DateTime: evaluationTimestamp,
 		Version:  evaluate.Version,
@@ -290,8 +290,8 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		return err
 	}
 
-	_ = metrics.WalkByScore(assessmentsPerModel, func(model string, assessment metrics.Assessments, score uint) error {
-		log.Printf("Evaluation score for %q (%q): %s", model, assessment.Category(totalScore).ID, assessment)
+	_ = assessmentsPerModel.WalkByScore(func(model model.Model, assessment metrics.Assessments, score uint) error {
+		log.Printf("Evaluation score for %q (%q): %s", model.ID(), assessment.Category(totalScore).ID, assessment)
 
 		return nil
 	})
--- a/cmd/eval-dev-quality/cmd/evaluate.go
+++ b/cmd/eval-dev-quality/cmd/evaluate.go
@@ -317,5 +317,17 @@ func writeCSVs(resultPath string, assessments report.AssessmentPerModelPerLangua
 		return pkgerrors.Wrap(err, "could not write models-summed.csv summary")
 	}
 
+	// Write the individual "language-summed.csv" containing the summary per model per language.
+	byLanguage := assessments.CollapseByLanguage()
+	for language, modelsByLanguage := range byLanguage {
+		csvByLanguage, err := report.GenerateCSV(modelsByLanguage)
+		if err != nil {
+			return pkgerrors.Wrap(err, "could not create "+language.ID()+"-summed.csv summary")
+		}
+		if err := os.WriteFile(filepath.Join(resultPath, language.ID()+"-summed.csv"), []byte(csvByLanguage), 0644); err != nil {
+			return pkgerrors.Wrap(err, "could not write "+language.ID()+"-summed.csv summary")
+		}
+	}
+
 	return nil
 }
--- a/cmd/eval-dev-quality/cmd/evaluate.go
+++ b/cmd/eval-dev-quality/cmd/evaluate.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	pkgerrors "github.com/pkg/errors"
 	"github.com/zimmski/osutil"
 	"golang.org/x/exp/maps"
 
@@ -252,14 +253,6 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		}
 	}
 
-	csv, err := report.GenerateCSV(assessments)
-	if err != nil {
-		log.Panicf("ERROR: could not create result summary: %s", err)
-	}
-	if err := os.WriteFile(filepath.Join(command.ResultPath, "evaluation.csv"), []byte(csv), 0644); err != nil {
-		log.Panicf("ERROR: could not write result summary: %s", err)
-	}
-
 	totalScore := uint(0)
 	// Set the total score to the number of evaluated languages if we are just checking the "plain" repositories since there is only one task to solve per language.
 	isOnlyPlainRepositories := true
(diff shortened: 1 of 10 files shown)


zimmski:
@Munsio @bauersimon when the PR takes too long we aim to early merge. Same rules as internally.

evaluate/report/markdown.go
```diff
@@ -34,7 +34,7 @@ type Markdown struct {
 	SVGPath string
 
 	// AssessmentPerModel holds
-	AssessmentPerModel map[string]metrics.Assessments
+	AssessmentPerModel AssessmentPerModel
```
bauersimon, line 37: Broken docstring here... could you please complete it :sweat_smile:

zimmski:
@Munsio please groom the issue with a task list. Planning as usual, do not jump in.

evaluate/report/collection.go
```diff
@@ -86,10 +86,13 @@ func (a AssessmentPerModelPerLanguagePerRepository) CollapseByModel() Assessment
...
 
 // WalkByScore walks the given assessment metrics by their score.
-func WalkByScore(assessmentsPerModel map[string]metrics.Assessments, function func(model string, assessment metrics.Assessments, score uint) error) error {
+func WalkByScore(assessmentsPerModel AssessmentPerModel, function func(model model.Model, assessment metrics.Assessments, score uint) error) error {
 	models := maps.Keys(assessmentsPerModel)
-	sort.Strings(models)
-	scores := make(map[string]uint, len(models))
+	sort.SliceStable(models, func(i, j int) bool {
+		return models[i].ID() < models[j].ID()
+	})
```
bauersimon, line 116: Would use the following to avoid dealing with indices:

```go
slices.SortStableFunc(models, func(a, b model.Model) int {
	return cmp.Compare(a.ID(), b.ID())
})
```

evaluate/report/collection.go
```diff
@@ -86,10 +86,13 @@ func (a AssessmentPerModelPerLanguagePerRepository) CollapseByModel() Assessment
 }
 
 // WalkByScore walks the given assessment metrics by their score.
-func WalkByScore(assessmentsPerModel map[string]metrics.Assessments, function func(model string, assessment metrics.Assessments, score uint) error) error {
+func WalkByScore(assessmentsPerModel AssessmentPerModel, function func(model model.Model, assessment metrics.Assessments, score uint) error) error {
```
bauersimon, line 112: How about making `WalkByScore` a method of `AssessmentPerModel`?

zimmski:
Reworded PR title. "Additional CSVs" is not enough to say what we are. Please hold to the same what-and-why pattern as internal. Need to write changelogs, and not having good titles/descriptions makes that much harder. Much faster that way.

evaluate/report/collection.go
```diff
@@ -20,6 +20,9 @@ type AssessmentPerModelPerLanguagePerRepository map[model.Model]map[language.Lan
 // AssessmentPerModel holds a collection of assessments per model.
 type AssessmentPerModel map[model.Model]metrics.Assessments
 
+// AssessmentPerLanguagePerModel holds a collection of assessments per language and model.
+type AssessmentPerLanguagePerModel map[language.Language]map[model.Model]metrics.Assessments
```
bauersimon, line 24: I would've changed the order to keep the model up front. It shouldn't really matter, right? Or is this something that Markus' hacky version explicitly did?

bauersimon: I guess it will only change the ordering in the CSV file in the end. And sorting by language first might be nice.

bauersimon (changes requested):
Awesome!!!

evaluate/report/collection.go
Munsio: Sounds good, will change

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -274,7 +274,7 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		totalScore = uint(len(languagesSelected))
 	}
 
-	assessmentsPerModel := assessments.Collapse()
+	assessmentsPerModel := assessments.CollapseByModel()
```
bauersimon, line 270: Would rename commit to "refactor, New types for storing assessments per model and per language"

evaluate/report/collection.go
```diff
@@ -14,6 +14,36 @@ import (
 	"github.com/symflower/eval-dev-quality/model"
 )
 
+// AssessmentPerLanguagePerModel holds a collection of assessments per language and model.
+type AssessmentPerLanguagePerModel map[language.Language]map[model.Model]metrics.Assessments
+
```
bauersimon: never used in this commit :(

should probably be part of the next one instead

evaluate/report/collection.go
```diff
@@ -67,17 +97,37 @@ func (a AssessmentPerModelPerLanguagePerRepository) Walk(function func(m model.M
...
+			assessments[l][m] = metrics.NewAssessments()
+		}
+
+		assessments[l][m].Add(a)
+
+		return nil
+	})
+
+	return assessments
+}
```
bauersimon, line 133: never used in this commit :(

should probably be part of the next one instead

evaluate/report/collection.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // AssessmentPerLanguagePerModel holds a collection of assessments per language and model.
-type AssessmentPerLanguagePerModel map[language.Language]map[model.Model]metrics.Assessments
+type AssessmentPerLanguagePerModel map[language.Language]AssessmentPerModel
```
bauersimon, line 18: This commit is only about language (since model worked before). Plz adapt the commit msg.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -296,5 +289,23 @@ func (command *Evaluate) Execute(args []string) (err error) {
...
 
+	if err := writeCSVs(command.ResultPath, assessments); err != nil {
+		log.Panicf("ERROR: %s", err)
+	}
+
+	return nil
+}
+
+// WriteCSVs writes the various CSV reports to disk.
+func writeCSVs(resultPath string, assessments report.AssessmentPerModelPerLanguagePerRepository) (err error) {
```
bauersimon, line 300: nitpick

this change does not really have anything to do with the newly introduced interface of this commit

evaluate/report/csv.go
```diff
@@ -59,3 +59,17 @@ func (a AssessmentPerModelPerLanguagePerRepository) Rows() (rows [][]string) {
...
+
+// Header returns the header description as a CSV row.
+func (a AssessmentPerModel) Header() (header []string) {
+	return []string{"model", "score"}
+}
+
+// Rows returns all data as CSV rows.
+func (a AssessmentPerModel) Rows() (rows [][]string) {
+	for model, assessment := range a {
+		rows = append(rows, []string{model.ID(), strconv.FormatUint(uint64(assessment.Score()), 10)})
```
bauersimon, line 71: I think it would be nice to still have all the metrics and not just the score. Just to have as much data as possible.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -317,5 +317,17 @@ func writeCSVs(resultPath string, assessments report.AssessmentPerModelPerLangua
 		return pkgerrors.Wrap(err, "could not write models-summed.csv summary")
 	}
 
+	// Write the individual "language-summed.csv" containing the summary per model per language.
+	byLanguage := assessments.CollapseByLanguage()
+	for language, modelsByLanguage := range byLanguage {
+		csvByLanguage, err := report.GenerateCSV(modelsByLanguage)
```
bauersimon, line 323: Is this commit missing something? I cannot find the header + row implementation of "by language"?

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: so should maybe go into the next one?

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: It doesn't need it as the language index is only for the different files. The assessment underneath is `AssessmentPerModel` and that has the formatting.

evaluate/report/csv.go
Munsio: I've based it upon what we do have now in the csvs, but sure can do that.

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: Well due to the change to an Interface I needed to revise how the `evaluation.csv` is written that is why it is in there.

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: I have no clue what commit you are currently talking about, Github doesn't link to the commit file but rather to all changes, we need to add the commit hash in such comments in the future for better tracing.

bauersimon (approved):
Beautiful

PR #95, opened 2024-05-06
Ollama tool installation

bauersimon:

PR #96, opened 2024-05-07
Support Ollama provider

bauersimon:
Early merger is #95.

### DIFF
--- a/cmd/eval-dev-quality/cmd/evaluate.go
+++ b/cmd/eval-dev-quality/cmd/evaluate.go
@@ -21,6 +21,7 @@ import (
 	"github.com/symflower/eval-dev-quality/log"
 	"github.com/symflower/eval-dev-quality/model"
 	"github.com/symflower/eval-dev-quality/provider"
+	_ "github.com/symflower/eval-dev-quality/provider/ollama"     // Register provider.
 	_ "github.com/symflower/eval-dev-quality/provider/openrouter" // Register provider.
 	_ "github.com/symflower/eval-dev-quality/provider/symflower"  // Register provider.
 	"github.com/symflower/eval-dev-quality/tools"
@@ -30,6 +31,10 @@ import (
 type Evaluate struct {
 	// InstallToolsPath determines where tools for the evaluation are installed.
 	InstallToolsPath string `long:"install-tools-path" description:"Install tools for the evaluation into this path."`
+	// OllamaBinaryPath overwrites the Ollama binary path.
+	OllamaBinaryPath string `long:"ollama-binary-path" description:"Overwrite the Ollama binary with this specific path instead of using a global one." env:"OLLAMA_BINARY_PATH"`
+	// OllamaURL overwrites the Ollama URL.
+	OllamaURL string `long:"ollama-url" description:"Overwrite the URL of the Ollama service." env:"OLLAMA_URL"`
 	// SymflowerBinaryPath overwrites the Symflower binary path.
 	SymflowerBinaryPath string `long:"symflower-binary-path" description:"Overwrite the Symflower binary with this specific path instead of installing and using a global one." env:"SYMFLOWER_BINARY_PATH"`
 
@@ -72,6 +77,27 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		return err
 	}
 	defer logClose()
+
+	// Check common options.
+	{
+		if command.InstallToolsPath == "" {
+			command.InstallToolsPath, err = tools.InstallPathDefault()
+			if err != nil {
+				log.Panicf("ERROR: %s", err)
+			}
+		}
+
+		if command.OllamaBinaryPath != "" {
+			tools.OllamaPath = command.OllamaBinaryPath
+		}
+		if command.OllamaURL != "" {
+			tools.OllamaURL = command.OllamaURL
+		}
+
+		if command.SymflowerBinaryPath != "" {
+			tools.SymflowerPath = command.SymflowerBinaryPath
+		}
+	}
 
 	// Gather languages.
 	languagesSelected := map[string]language.Language{}
@@ -130,6 +156,8 @@ func (command *Evaluate) Execute(args []string) (err error) {
 	{
 		models := map[string]model.Model{}
 		for _, p := range provider.Providers {
+			log.Printf("Checking provider %q for models", p.ID())
+
 			if t, ok := p.(provider.InjectToken); ok {
 				token, ok := command.ProviderTokens[p.ID()]
 				if ok {
@@ -137,9 +165,23 @@ func (command *Evaluate) Execute(args []string) (err error) {
 				}
 			}
 			if err := p.Available(log); err != nil {
-				log.Printf("skipping unavailable provider %q cause: %s", p.ID(), err)
+				log.Printf("Skipping unavailable provider %q cause: %s", p.ID(), err)
 
 				continue
+			}
+
+			// Start services of providers.
+			if service, ok := p.(provider.Service); ok {
+				log.Printf("Starting services for provider %q", p.ID())
+				shutdown, err := service.Start(log)
+				if err != nil {
+					log.Panicf("ERROR: could not start services for provider %q: %s", p, err)
+				}
+				defer func() {
+					if err := shutdown(); err != nil {
+						log.Panicf("ERROR: could not shutdown services of provider %q: %s", p, err)
+					}
+				}()
 			}
 
 			ms, err := p.Models()
@@ -177,21 +219,8 @@ func (command *Evaluate) Execute(args []string) (err error) {
 	}
 
 	// Install required tools for the basic evaluation.
-	{
-		if command.SymflowerBinaryPath != "" {
-			tools.SymflowerPath = command.SymflowerBinaryPath
-		}
-
-		if command.InstallToolsPath == "" {
-			command.InstallToolsPath, err = tools.InstallPathDefault()
-			if err != nil {
-				log.Panicf("ERROR: %s", err)
-			}
-		}
-
-		if err := tools.InstallEvaluation(log, command.InstallToolsPath); err != nil {
-			log.Panicf("ERROR: %s", err)
-		}
+	if err := tools.InstallEvaluation(log, command.InstallToolsPath); err != nil {
+		log.Panicf("ERROR: %s", err)
 	}
 
 	// Check that models and languages can be evaluated by executing the "plain" repositories.
--- a/cmd/eval-dev-quality/cmd/evaluate_test.go
+++ b/cmd/eval-dev-quality/cmd/evaluate_test.go
@@ -21,6 +21,7 @@ import (
 	modeltesting "github.com/symflower/eval-dev-quality/model/testing"
 	"github.com/symflower/eval-dev-quality/provider"
 	providertesting "github.com/symflower/eval-dev-quality/provider/testing"
+	"github.com/symflower/eval-dev-quality/tools"
 )
 
 // validateReportLinks checks if the Markdown report data contains all the links to other relevant report files.
@@ -52,7 +53,7 @@ func TestEvaluateExecute(t *testing.T) {
 
 		ExpectedOutputValidate func(t *testing.T, output string, resultPath string)
 		ExpectedResultFiles    map[string]func(t *testing.T, filePath string, data string)
-		ExpectedPanic          string
+		ExpectedPanicContains  string
 	}
 
 	validate := func(t *testing.T, tc *testCase) {
@@ -79,14 +80,24 @@ func TestEvaluateExecute(t *testing.T) {
 				"--testdata", filepath.Join("..", "..", "..", "testdata"),
 			}, tc.Arguments...)
 
-			if tc.ExpectedPanic == "" {
+			if tc.ExpectedPanicContains == "" {
 				assert.NotPanics(t, func() {
 					Execute(logger, arguments)
 				})
 			} else {
-				assert.PanicsWithValue(t, tc.ExpectedPanic, func() {
+				didPanic := true
+				var recovered any
+				func() {
+					defer func() {
+						recovered = recover()
+					}()
+
 					Execute(logger, arguments)
-				})
+
+					didPanic = false
+				}()
+				assert.True(t, didPanic)
+				assert.Contains(t, recovered, tc.ExpectedPanicContains)
 			}
 
 			if tc.ExpectedOutputValidate != nil {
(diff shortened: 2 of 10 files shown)


bauersimon:
TODO: tests pass when targeted manually but running `make test` fails nearly all the tests with the reason that the local Ollama server is unreachable. I guess this is because we always use the default port and the tests run in parallel, but running them with the `-p 1` flag does not work either.

bauersimon:
```sh
#!/bin/bash

# Checkout and install Ollama as part of tools.
git fresh-checkout ollama
make install-all

# Ollama server instance which manages all models.
~/.eval-dev-quality/bin/ollama serve &
ollamaPID=$!

# Wait a second for the server to start.
sleep 3

# Pull some models (check https://ollama.com/library for more but watch out, many are several GBs large).
~/.eval-dev-quality/bin/ollama pull qwen:0.5b
~/.eval-dev-quality/bin/ollama pull tinyllama:latest

# Kill the server, or don't cause our eval sould either start a new one or reuse the existing one.
kill $ollamaPID

# Wait a second for the server to stop.
sleep 3

# Run the eval with only the models we pulled.
eval-dev-quality evaluate --model ollama/qwen:0.5b --model ollama/tinyllama:latest

# Kill the server just in case we left it running.
trap 'kill $(jobs -p)' EXIT
```

@Munsio @ahumenberger

Munsio:
I've run the bash script and it gave me the following output:
https://gist.github.com/Munsio/8dcae3ff9d58c0ccf25401287ea42619

util/exec.go
```diff
@@ -21,16 +23,17 @@ type Command struct {
...
 
 	var writer bytesutil.SynchronizedBuffer
-	c := exec.Command(command.Command[0], command.Command[1:]...)
+	c := exec.CommandContext(ctx, command.Command[0], command.Command[1:]...)
 	if command.Directory != "" {
 		c.Dir = command.Directory
 	}
 	c.Stdout = io.MultiWriter(logger.Writer(), &writer)
 	c.Stderr = c.Stdout
+	c.WaitDelay = 3 * time.Second
```
zimmski, line 36: I assume this is for Ollama. Please always document something like this: extra options / settings. 3 is also a very curious value.

util/exec.go
```diff
@@ -21,16 +23,17 @@ type Command struct {
 }
 
 // CommandWithResult executes a command and returns its output, while printing the same output to the given logger.
-func CommandWithResult(logger *log.Logger, command *Command) (output string, err error) {
+func CommandWithResult(ctx context.Context, logger *log.Logger, command *Command) (output string, err error) {
```
zimmski, line 26: Please always merge a commit like this earlier. It has now a conflict, 100% avoidable. Nothing special to review.

util/exec.go
zimmski: @bauersimon still putting this in the commit. but not commenting, so please do a follow up PR commenting it.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -45,8 +48,8 @@ func TestEvaluateExecute(t *testing.T) {
 	type testCase struct {
 		Name string
 
-		Before func(t *testing.T, resultPath string)
-		After  func(t *testing.T, resultPath string)
+		Before func(t *testing.T, logger *log.Logger, resultPath string)
```
zimmski, line 51: This is a refactoring. Please pull such changes out.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -93,11 +96,17 @@ func TestEvaluateExecute(t *testing.T) {
...
 			}
 
-			actualResultFiles, err := osutil.FilesRecursive(temporaryPath)
+			actualResultFilesRaw, err := osutil.FilesRecursive(temporaryPath)
 			require.NoError(t, err)
-			for i, p := range actualResultFiles {
-				actualResultFiles[i], err = filepath.Rel(temporaryPath, p)
+			actualResultFiles := make([]string, 0, len(actualResultFilesRaw))
+			for _, p := range actualResultFilesRaw {
+				if strings.Contains(p, "TESTIGNORE") {
```
zimmski, line 105: What is that?

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -30,8 +31,10 @@ import (
 type Evaluate struct {
 	// InstallToolsPath determines where tools for the evaluation are installed.
 	InstallToolsPath string `long:"install-tools-path" description:"Install tools for the evaluation into this path."`
-	// SymflowerBinaryPath overwites the Symflower binary path.
+	// SymflowerBinaryPath overwrites the Symflower binary path.
 	SymflowerBinaryPath string `long:"symflower-binary-path" description:"Overwrite the Symflower binary with this specific path instead of installing and using a global one." env:"SYMFLOWER_BINARY_PATH"`
+	//OllamaBinaryPath overwrites the Ollama binary path.
+	OllamaBinaryPath string `long:"ollama-binary-path" description:"Overwrite the Ollama binary with this specific path instead of installing and using a global one." env:"OLLAMA_BINARY_PATH"`
```
zimmski, line 37: Sort

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -166,6 +188,8 @@ func (command *Evaluate) Execute(args []string) (err error) {
 		for _, modelID := range command.Models {
 			modelsSelected[modelID] = models[modelID]
 		}
+
+		// TODO Download models in case the provider supports it. https://github.com/symflower/eval-dev-quality/issues/100
```
zimmski, line 192: Will remove the TODO. It is in the issue. no need to clutter. Not a priority TODO.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -142,6 +149,21 @@ func (command *Evaluate) Execute(args []string) (err error) {
 				continue
 			}
 
+			// Startup any providers that need it.
```
zimmski, line 152: Vague comment. "that need it"

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -73,6 +76,10 @@ func (command *Evaluate) Execute(args []string) (err error) {
 	}
 	defer logClose()
 
+	if command.OllamaBinaryPath != "" {
```
zimmski, line 79: Let's move other option validations here too.

tools/ollama_test.go
```diff
@@ -2,8 +2,27 @@ package tools
...
+	defer func() {
+		if t.Failed() {
+			t.Log(buffer.String())
+		}
+	}()
+
+	shutdown, err := StartOllamaServer(logger, OllamaPath)
+	assert.NoError(t, err)
+
+	time.Sleep(3 * time.Second)
```
zimmski, line 29: There should be a check.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -60,6 +60,8 @@ func TestEvaluateExecute(t *testing.T) {
 
 	validate := func(t *testing.T, tc *testCase) {
 		t.Run(tc.Name, func(t *testing.T) {
+			providertesting.OllamaInstanceTest(t)
```
zimmski, line 63: This commit should not be a workaround. It is fundamental.

provider/provider.go
```diff
@@ -47,3 +47,10 @@ type Query interface {
 	// Query queries the provider with the given model name.
 	Query(ctx context.Context, modelIdentifier string, promptText string) (response string, err error)
 }
+
+// Background is a provider with background services.
+type Background interface {
```
zimmski, line 52: Will rename this to "Service" sure. it runs in the background, but what we basically want to know is if this provider requires a service or not.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -305,6 +314,8 @@ func TestEvaluateExecute(t *testing.T) {
 	})
 	t.Run("Model filter", func(t *testing.T) {
 		t.Run("openrouter.ai", func(t *testing.T) {
+			t.Setenv("OLLAMA_MODELS", t.TempDir()) // Set the Ollama model directory to something else so even if the user executing these tests has Ollama installed, we don't query their models.
```
zimmski, line 317: I think, we want that. I do not want to query all models again that are run in the tests. So by default, i think we should use the cached models. Removing this.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -322,13 +333,53 @@ func TestEvaluateExecute(t *testing.T) {
 				ExpectedPanic: "ERROR: model openrouter/auto does not exist. Valid models are: symflower/symbolic-execution",
 			})
 		})
+		t.Run("Ollama", func(t *testing.T) {
+			validate(t, &testCase{
+				Name: "Pulled Model",
+
+				Before: func(t *testing.T, logger *log.Logger, resultPath string) {
+					t.Setenv("OLLAMA_MODELS", filepath.Join(resultPath, "TESTIGNORE-ollama-models"))
+					shutdown, err := tools.StartOllamaServer(logger, tools.OllamaPath)
```
zimmski, line 342: Will change this to be a service that can be used.

cmd/eval-dev-quality/cmd/evaluate_test.go
zimmski: Ah, this is to ignore the ollama model directory. I think it is better to just add subdirectories to the temporary directory, because then you always have a subdirectory that is clean of anything else.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -322,13 +333,53 @@ func TestEvaluateExecute(t *testing.T) {
...
+					require.NoError(t, err)
+
+					_, err = util.CommandWithResult(context.Background(), logger, &util.Command{
```
zimmski, line 345: Should be a service method.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -322,13 +333,53 @@ func TestEvaluateExecute(t *testing.T) {
...
+						Command: []string{
+							tools.OllamaPath,
+							"pull", "qwen:0.5b",
+						},
+					})
+					require.NoError(t, err)
+					require.NoError(t, shutdown())
```
zimmski, line 352: I think the best action for the service is to make it not only dependent on the binary, but first of all on the host URL. Reason is that we can then use the service right away over HTTP.

tools/ollama.go
```diff
@@ -87,3 +105,155 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
...
+	ollamaProcessStatesLock sync.Mutex
+)
+
+type ollamaProcessState struct {
+	// url holds the URL of the Ollama process.
+	url string
+	// connectionCount holds the amount of connections.
+	// When the counter hits `0`, the process will be removed.
+	connectionCount uint
+	// cleanup holds functio to clean up the Ollama process.
```
bauersimon, line 125: typo

tools/ollama.go
```diff
@@ -87,3 +105,155 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
...
+		return func() (err error) {
+			ollamaProcessStateRemove(url)
+
+			return nil
+		}, nil
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	logger.Printf("Starting new Ollama service on %q using %s", url, binaryPath)
```
bauersimon, line 208: The url is never set at the server command. I believe it must be an env variable so our exec util should be adapted to support envs?

util/exec.go
```diff
@@ -19,6 +19,9 @@ type Command struct {
 
 	// Directory defines the directory the execution should run in, without changing the working directory of the caller.
 	Directory string
+	// Env defines all environment variables for the executed command.
+	// If HOME and PATH are not included, they will be set automatically.
+	Env map[string]string
```
bauersimon, line 26: Would say that the command inherits all env variables unless overwritten in the map.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -314,14 +325,63 @@ func TestEvaluateExecute(t *testing.T) {
...
+						var err error
+						shutdown, err = tools.OllamaStart(logger, tools.OllamaPath, tools.OllamaURL)
+						require.NoError(t, err)
+
+						require.NoError(t, tools.OllamaPull(logger, tools.OllamaPath, tools.OllamaURL, "qwen:0.5b"))
+					},
+					After: func(t *testing.T, logger *log.Logger, resultPath string) {
+						require.NoError(t, shutdown())
+						shutdown = nil
+					},
```
bauersimon, line 361: Could probably get rid of this shutdown and rely on the `defer` one outside of the `valdidate`... Not really a fan of doubling this up.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -314,14 +325,63 @@ func TestEvaluateExecute(t *testing.T) {
...
+
+			{
+				var shutdown func() (err error)
+				defer func() {
+					if shutdown != nil {
+						require.NoError(t, shutdown())
+					}
+				}()
+				validate(t, &testCase{
+					Name: "Pulled Model",
```
bauersimon, line 349: This test is having a running server in the background. We could also do one where the eval itself needs to start the server. But since we anyways have ony one single instance... I guess that is pointless.

provider/ollama/ollama_test.go
```diff
@@ -0,0 +1,152 @@
...
+					defer cleanup()
+				}
+			}
+
+			provider.SetURL(url)
+
+			shutdown, actualErr := provider.Start(logger)
+			assert.NoError(t, actualErr)
+
+			// TODO do a check.
```
bauersimon, line 52: Leftover TODO.

tools/ollama.go
```diff
@@ -63,11 +71,11 @@ func (*ollama) CheckVersion(logger *log.Logger, binaryPath string) (err error) {
 
 	versionPresent := "v" + m[1]
 	if !semver.IsValid(versionPresent) {
-		pkgerrors.New(fmt.Sprintf("invalid semantic version string %q", versionPresent))
+		return pkgerrors.New(fmt.Sprintf("invalid semantic version string %q", versionPresent))
```
bauersimon, line 74: woops :see_no_evil:

provider/ollama/ollama.go
```diff
@@ -0,0 +1,125 @@
...
+	return &Provider{
+		binaryPath: tools.OllamaPath,
+		url:        tools.OllamaURL,
+	}
+}
+
+// SetBinaryPath sets the binary file path of the Ollama binary.
+func (p *Provider) SetBinaryPath(binaryPath string) {
+	p.binaryPath = binaryPath
+}
```
bauersimon, line 42: Unused function.

provider/ollama/ollama.go
```diff
@@ -0,0 +1,125 @@
...
+
+// SetURL sets the URL to the Ollama service.
+func (p *Provider) SetURL(url string) {
+	p.url = url
+}
```
bauersimon, line 47: Only used in tests, so I'd say could also directly modify the struct content and get rid of this function.

tools/ollama.go
```diff
@@ -120,3 +128,203 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
...
+	ollamaProcessStatesLock sync.Mutex
+)
+
+type ollamaProcessState struct {
+	// url holds the URL of the Ollama process.
+	url string
+	// connectionCount holds the amount of connections.
+	// When the counter hits `0`, the process will be removed.
+	connectionCount uint
+	// cleanup holds function to clean up the Ollama process.
```
bauersimon, line 148: Typo/Gramma

tools/ollama.go
```diff
@@ -120,3 +128,203 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
...
+	state, ok := ollamaProcessStates[url]
+	if !ok {
+		state = &ollamaProcessState{
+			url: url,
+		}
+		ollamaProcessStates[url] = state
+	}
+	if cleanup != nil {
+		state.cleanup = cleanup // A process can be started afterwards as well, so make sure it is always cleaned up.
+	}
```
bauersimon, line 164: corner case: if the state already existed and had a cleanup function, that function will be overwritten

tools/ollama.go
```diff
@@ -120,3 +128,203 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
...
+		} else if response.StatusCode != http.StatusOK {
+			return pkgerrors.WithStack(fmt.Errorf("unexpected status code %d", response.StatusCode))
+		}
+
+		bodyBytes, err := io.ReadAll(response.Body)
+		if err != nil {
+			return pkgerrors.WithStack(err)
+		} else if string(bodyBytes) != "Ollama is running" {
+			return pkgerrors.WithStack(err)
+		}
```
bauersimon, line 202: Second case will return `err` which is `nil` in that case.

tools/ollama.go
```diff
@@ -120,3 +128,203 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
...
+
+		return nil
+	}, retry.Delay(time.Second), retry.Attempts(3)); err != nil {
+		return pkgerrors.WithStack(pkgerrors.WithMessage(err, "unable to start Ollama server"))
```
bauersimon, line 206: Stale error msg: should be something like, "could not detect running Ollama instance"

util/exec.go
```diff
@@ -45,6 +47,8 @@ func CommandWithResult(ctx context.Context, logger *log.Logger, command *Command
 	c.Stdout = io.MultiWriter(logger.Writer(), &writer)
 	c.Stderr = c.Stdout
 
+	c.WaitDelay = 3 * time.Second // Some binary do not like to be killed, e.g. "ollama", so we kill them after some time automatically.
```
bauersimon, line 50: Typo: binaries... but I like it the way it is as well.

util/exec.go
zimmski: Wups. copy&pasta error

PR #99, opened 2024-05-08
Sort map by model before creating the CSV output

Munsio:


bauersimon:
awesome - could you maybe add the `untilfail go test ...` command of the test case in the commit message for reference?

bauersimon (approved):

PR #102, opened 2024-05-08
Support MacOS

zimmski:


zimmski:
@Munsio merging this right away but tagging you to take a look at the CI config. I am sure we can do better. There is nothing to it, and Symflower makes this extremely easy to support. We did a nice job:

PR #103, opened 2024-05-08
Support Windows

zimmski:


zimmski:
@Munsio merging this right away to move forward with Ollama. This was not straight forward. Windows is massively incompatible with me.  Please take a look at the commits. I think we could do better with that PR, but i am  desperate to move on and tests should be green now (btw Maven, again, flaky... who knew?)

PR #104, opened 2024-05-11
fix, Let Windows CI error on failures, and inject API tokens only once per provider

zimmski:


zimmski:
Early merger of https://github.com/symflower/eval-dev-quality/pull/95. Will merge right away.

PR #110, opened 2024-05-13
Don't log that we exclude models because we never do

bauersimon:
Was not adapted in f6965c583195c2c86d9969ed67038a0eacfe22c5


bauersimon:
nvmind actually... we do exclude them... just later on... confusing now that we have this log here...

bauersimon:
misunderstanding

PR #114, opened 2024-05-14
 Use Git to avoid copying the repository on each model run

Munsio:


evaluate/repository.go
```diff
@@ -77,3 +78,56 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
...
+
+// TemporaryRepository creates a temporary repository and initializes a git repo in it.
+func TemporaryRepository(testDataPath string, repositoryPath string) (temporaryRepositoryPath string, err error) {
+	temporaryPath, err := os.MkdirTemp("", "eval-dev-quality")
+	if err != nil {
+		return "", pkgerrors.WithStack(err)
+	}
+
+	dataPath := filepath.Join(testDataPath, repositoryPath)
+	temporaryRepositoryPath = filepath.Join(temporaryPath, filepath.Base(repositoryPath))
```
bauersimon, line 70: Do you really mean `filepath.Base` or should that be `filepath.Dir`? Just making sure because I made such a mistake as well.

evaluate/repository.go
```diff
@@ -77,3 +78,56 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
 
 	return repositoryAssessment, problems, nil
 }
+
+// TemporaryRepository creates a temporary repository and initializes a git repo in it.
+func TemporaryRepository(testDataPath string, repositoryPath string) (temporaryRepositoryPath string, err error) {
```
bauersimon, line 63: Why do we have `testDataPath` and `repositoryPath` separate? We concat them anyways into `dataPath`  within this function.

evaluate/repository.go
```diff
@@ -77,3 +78,56 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
...
+	temporaryPath, err := os.MkdirTemp("", "eval-dev-quality")
```
bauersimon, line 64: This temporary directory is never ever cleaned up: https://pkg.go.dev/os#MkdirTemp

How about returning a cleanup function

evaluate/repository_test.go
```diff
@@ -93,3 +93,77 @@ func TestRepository(t *testing.T) {
...
+		t.Run(tc.Name, func(t *testing.T) {
+			temporaryRepositoryPath, err := TemporaryRepository(tc.TestDataPath, tc.RepositoryPath)
+			assert.NoError(t, err)
+			defer func() {
+				assert.NoError(t, os.RemoveAll(temporaryRepositoryPath))
+			}()
+
+			// Create a file
+			tmpFile := filepath.Join(temporaryRepositoryPath, "foo")
+			assert.NoError(t, os.WriteFile(tmpFile, []byte("foo"), 0600))
```
bauersimon, line 151: Nitpick: this is somewhat testcase specific and would probably be better within the `testCase` instead of within `validate`.

evaluate/repository.go
```diff
@@ -95,16 +75,16 @@ func TemporaryRepository(testDataPath string, repositoryPath string) (temporaryR
 	// Add git repository and commit changes.
 	err = osutil.Chdir(temporaryPath, func() error {
 		cmd := exec.Command("git", "init")
-		if err := cmd.Run(); err != nil {
-			return pkgerrors.Wrap(err, "unable to initialize git repository")
+		if out, err := cmd.CombinedOutput(); err != nil {
+			return pkgerrors.Wrap(err, fmt.Sprintf("%s - %s", "unable to initialize git repository", out))
```
bauersimon, line 79: Could put this in separate commit.

evaluate/repository.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 // Repository evaluate a repository with the given model and language.
-func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language, testDataPath string, repositoryPath string) (repositoryAssessment metrics.Assessments, problems []error, err error) {
+func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language, temporaryRepositoryPath string, repositoryPath string) (repositoryAssessment metrics.Assessments, problems []error, err error) {
```
bauersimon, line 19: I think the names `temporaryRepositoryPath` and `repositoryPath` might be confusing... How about keeping `testDataPath` and renaming `repositoryPath` to `repositoryName` since it is only used for logging as far as I can see.

bauersimon (dismissed):

bauersimon:
watch out - will rebase

zimmski:
Will add a workaorund for right now: copying for every generation. Need to have the clean up in right now, and this PR seems not ready to move on.

evaluate/repository.go
Munsio: I want the last element of `repositoryPath` as it is the "identifier" for the repository inside the temporaryPath. We could also use the full repositoryPath but that seems to much imho.

evaluate/repository.go
```diff
@@ -80,3 +83,86 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
 
 	return repositoryAssessment, problems, nil
 }
+
+// TemporaryRepository creates a temporary repository and initializes a git repo in it.
+func TemporaryRepository(logger *log.Logger, dataPath string) (temporaryRepositoryPath string, cleanup func(), err error) {
+	temporaryPath, err := os.MkdirTemp("", "eval-dev-quality")
```
ruiAzevedo19, line 69: Why not handling the error right way and only after defining the `cleanup` function?

evaluate/repository.go
```diff
@@ -80,3 +83,86 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
...
+		Directory: temporaryRepositoryPath,
+	})
+	if err != nil {
+		return "", cleanup, pkgerrors.WithStack(pkgerrors.Wrap(err, fmt.Sprintf("%s - %s", "unable to commit", out)))
+	}
+
+	return temporaryRepositoryPath, cleanup, nil
+}
+
+// ResetTemporaryRepository resets a temporary repository to back to its "initial" commit.
```
ruiAzevedo19, line 132: `repository to back to` -> `repository back to`

ruiAzevedo19 (dismissed):

evaluate/repository.go
Munsio: ah nice, that was a ctrl+c/ctrl+v mistake of course.

evaluate/repository_test.go
```diff
@@ -39,7 +39,10 @@ func TestRepository(t *testing.T) {
 			temporaryPath := t.TempDir()
 
 			_, logger := log.Buffer()
-			actualRepositoryAssessment, actualProblems, actualErr := Repository(logger, temporaryPath, tc.Model, tc.Language, tc.TestDataPath, tc.RepositoryPath)
+			temporaryRepositoryPath, cleanup, actualErr := TemporaryRepository(logger, filepath.Join(tc.TestDataPath, tc.RepositoryPath))
```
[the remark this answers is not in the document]
Munsio: Also copy paste error

Munsio:
@ruiAzevedo19 I removed the regex test from windows so we can merge this one. Will do a follow-up PR to find a fix for it.

evaluate/repository.go
```diff
@@ -80,3 +83,86 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
...
+
+	cleanup = func() {
+		if e := os.RemoveAll(temporaryPath); e != nil {
+			if err != nil {
+				err = errors.Join(err, pkgerrors.WithStack(e))
+			} else {
+				err = pkgerrors.WithStack(e)
+			}
+		}
+	}
```
ruiAzevedo19, line 82: Since we are returning the error above, does these `err` checks still apply? Or do the errors below play a role here?

Munsio: The `err` may be set inside the function where the cleanup function is getting deferred. So the `err` does not necessary reference the scope where the function is declared in.

PR #118, opened 2024-05-15
Ollama version check and update if version is outdated

### LINKED ISSUE #117: Fixed Ollama version
We currently allow any Ollama version but we want a fixed one, similar to Symflower.

bauersimon:
Closes #117


Munsio (approved):

tools/ollama.go
```diff
@@ -69,7 +102,7 @@ func (*ollama) Install(logger *log.Logger, installPath string) (err error) {
 		return pkgerrors.WithStack(fmt.Errorf("unsupported architecture %s", a))
 	}
 
-	ollamaDownloadURL := "https://ollama.com/download/ollama-linux-" + architectureIdentifier
+	ollamaDownloadURL := "https://ollama.com/download/ollama-linux-" + architectureIdentifier + "?version=" + ollamaVersion
```
Munsio, line 105: Not sure if we should move to the github release page with version instead.

Munsio (approved):

tools/ollama.go
bauersimon: Also possible. that here is just the official download location that their installer script uses.

bauersimon: I guess we can trust more in the uptime of GH directly.

tools/ollama.go
```diff
@@ -42,11 +46,40 @@ func (*ollama) BinaryPath() string {
...
+	if m == nil {
+		return pkgerrors.WithStack(pkgerrors.WithMessage(errors.New("cannot find version"), output))
+	}
+
+	versionPresent := "v" + m[1]
+	if !semver.IsValid(versionPresent) {
+		pkgerrors.New(fmt.Sprintf("invalid semantic version string %q", versionPresent))
+	}
+	versionRequired := "v" + ollamaVersion
+	if !semver.IsValid(versionPresent) {
```
zimmski, line 69: I guess this should be versionRequired ?

tools/ollama.go
```diff
@@ -42,11 +46,40 @@ func (*ollama) BinaryPath() string {
...
+		pkgerrors.New(fmt.Sprintf("invalid semantic version string %q", versionRequired))
+	}
+
+	// Binary is installed in a compatible version.
```
zimmski, line 73: That is a stement, should be behavioral.

zimmski:
Crazy. when there is a Ollama service running, the version message is different.

PR #120, opened 2024-05-16
Option to execute multiple runs non-interleaved

bauersimon:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -48,6 +48,8 @@ type Evaluate struct {
 	ResultPath string `long:"result-path" description:"Directory path where results should be written to. The placeholder \"%datetime%\" can be used for the current date and time." default:"evaluation-%datetime%"`
 	// Runs holds the number of runs to perform.
 	Runs uint `long:"runs" description:"Number of runs to perform." default:"1"`
+	// RunsInterleavedDisabled indicates that interleaved runs are disabled.
+	RunsInterleavedDisabled bool `long:"disable-interleaved" description:"Multiple runs are performed in an interleaved fashion to avoid caching of model responses. Disabling this behavior queries the same model repeatedly."`
```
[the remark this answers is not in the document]
Munsio: By setting the default value either to true/false the behavior explanation is still the same for the user.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -58,6 +60,22 @@ type Evaluate struct {
...
 
+func (command *Evaluate) runsAtLanguageLevel() uint {
+	if command.RunsInterleavedDisabled {
+		return 1
+	}
+
+	return command.Runs
+}
+
+func (command *Evaluate) runsAtModelLevel() uint {
```
Munsio, line 71: Same for documentation.

cmd/eval-dev-quality/cmd/evaluate.go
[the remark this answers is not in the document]
Munsio: I know you want to have `--disable-interleave` as flag but imho it is also better for automation to just have `runs-interleaved=true/false`.

Munsio (dismissed):

cmd/eval-dev-quality/cmd/evaluate.go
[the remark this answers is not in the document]
bauersimon: I cannot set the default value to true: "Could not add arguments group: boolean flag `runs-interleaved' may not have default values, they always default to `false' and can only be turned on" so I guess to work around that we need to make it a string and do the parsing ourselves? :see_no_evil:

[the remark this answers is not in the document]
Munsio: No the only thing that I could think of would be `run-model-sequential` or something like that. So using the opposite of `interleaved`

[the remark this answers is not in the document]
bauersimon: I like that one.

bauersimon:
Oh nooo. @Munsio is off today. But he requested changes which I did. But I cannot un-assign him. So this is basically un-mergeable until he is back :face_exhaling:  WHYYY? @zimmski

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -48,6 +48,8 @@ type Evaluate struct {
 	ResultPath string `long:"result-path" description:"Directory path where results should be written to. The placeholder \"%datetime%\" can be used for the current date and time." default:"evaluation-%datetime%"`
 	// Runs holds the number of runs to perform.
 	Runs uint `long:"runs" description:"Number of runs to perform." default:"1"`
+	// RunsSequential indicates that interleaved runs are disabled and runs are performed sequentially.
```
zimmski, line 51: Taking this PR.

zimmski: Nope. too much other stuff.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -576,6 +576,60 @@ func TestEvaluateExecute(t *testing.T) {
...
+				"golang-summed.csv": nil,
+				"models-summed.csv": nil,
+				"README.md":         nil,
+				filepath.Join("symflower_symbolic-execution", "golang", "golang", "plain.log"): func(t *testing.T, filePath, data string) {
+					assert.Equal(t, 3, strings.Count(data, `Evaluating model "symflower/symbolic-execution"`))
+				},
+			},
+		})
+		validate(t, &testCase{
+			Name: "Sequential",
```
zimmski, line 591: Test, that this is sequentiel.

Munsio: The log output validation is different between `interleaved` and `sequential` so we are testing already that the execution is truthfully `sequential`

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: Done

ruiAzevedo19:
@Munsio Review please

evaluate/evaluate.go
```diff
@@ -75,16 +91,26 @@ func Evaluate(ctx *Context) (assessments report.AssessmentPerModelPerLanguagePer
...
-					}
-					if len(ps) > 0 {
-						ctx.Log.Printf("Model %q was not able to solve the %q repository for language %q: %+v", modelID, repositoryPath, languageID, ps)
-						problemsPerModel[modelID] = append(problemsPerModel[modelID], ps...)
-					} else {
-						modelSucceededBasicChecksOfLanguage[model][language] = true
+					for rm := uint(0); rm < ctx.runsAtModelLevel(); rm++ {
+						if err := ResetTemporaryRepository(ctx.Log, temporaryRepositoryPath); err != nil {
+							ctx.Log.Panicf("ERROR: unable to reset temporary repository path: %s", err)
+						}
```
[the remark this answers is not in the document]
Munsio: That is okay, we need to reset the repository before each model run.

evaluate/evaluate.go
```diff
@@ -46,9 +66,9 @@ func Evaluate(ctx *Context) (assessments report.AssessmentPerModelPerLanguagePer
 	assessments = report.NewAssessmentPerModelPerLanguagePerRepository(ctx.Models, ctx.Languages, ctx.RepositoryPaths)
 	problemsPerModel := map[string][]error{}
 	{
-		for r := uint(0); r < ctx.Runs; r++ {
-			if ctx.Runs > 1 {
-				ctx.Log.Printf("Run %d/%d", r+1, ctx.Runs)
+		for rl := uint(0); rl < ctx.runsAtLanguageLevel(); rl++ {
```
Munsio, line 69: I guess I found a bug with the `TemporaryRepository`. We want to copy the repository for each language only 1 time but currently we are doing it on every run. So 3 runs = 3 x language copies.

We need to iterate over the languages before this loop and create the temporaryRepositoryPaths. We then need to store them inside a map and use the `repositoryPath` as index so we can fetch the correct temporary one during the model run.
The cleanup functions also need to be taken care of in a slice/map and run after everything is done.

Munsio: This is also true for the loop below where we are not using the `RepositoryPlainName`

Munsio (changes requested):

evaluate/evaluate.go
```diff
@@ -45,28 +65,39 @@ func Evaluate(ctx *Context) (assessments report.AssessmentPerModelPerLanguagePer
...
 	{
-		for r := uint(0); r < ctx.Runs; r++ {
-			if ctx.Runs > 1 {
-				ctx.Log.Printf("Run %d/%d", r+1, ctx.Runs)
+		// Create temporary repositories for each language so the repository is copied only once per language.
+		temporaryRepositories := map[string]*temporaryRepository{}
+		for _, language := range ctx.Languages {
+			repositoryPath := filepath.Join(language.ID(), RepositoryPlainName)
+			temporaryRepositoryPath, cleanup, err := TemporaryRepository(ctx.Log, filepath.Join(ctx.TestdataPath, repositoryPath))
+			if err != nil {
```
Munsio, line 80: I guess I got a little over board with the `cleanup` - you could just `defer` the cleanup after the error check and it should be alright.

Then you do not have to introduce a new type but rather a map[string]string

Munsio (changes requested):

evaluate/evaluate.go
```diff
@@ -75,20 +100,30 @@ func Evaluate(ctx *Context) (assessments report.AssessmentPerModelPerLanguagePer
...
-					if err != nil {
-						ps = append(ps, err)
-					}
-					if len(ps) > 0 {
-						ctx.Log.Printf("Model %q was not able to solve the %q repository for language %q: %+v", modelID, repositoryPath, languageID, ps)
-						problemsPerModel[modelID] = append(problemsPerModel[modelID], ps...)
-					} else {
-						modelSucceededBasicChecksOfLanguage[model][language] = true
+					for rm := uint(0); rm < ctx.runsAtModelLevel(); rm++ {
+						temporaryRepositoryPath := temporaryRepositories[repositoryPath]
```
Munsio, line 104: move that up to line 90 to avoid unnecessary reassignments.

evaluate/evaluate.go
```diff
@@ -126,33 +181,34 @@ func Evaluate(ctx *Context) (assessments report.AssessmentPerModelPerLanguagePer
...
 						continue
 					}
 
-					assessment, ps, err := Repository(ctx.Log, ctx.ResultPath, model, language, temporaryRepositoryPath, repositoryPath)
-					assessments[model][language][repositoryPath].Add(assessment)
-					problemsPerModel[modelID] = append(problemsPerModel[modelID], ps...)
-					if err != nil {
-						ctx.Log.Printf("ERROR: Model %q encountered a hard error for language %q, repository %q: %+v", modelID, languageID, repositoryPath, err)
+					for rm := uint(0); rm < ctx.runsAtModelLevel(); rm++ {
+						temporaryRepositoryPath := temporaryRepositories[repositoryPath]
```
Munsio, line 194: Here too move the assignment up to line 173

Munsio (changes requested):

Munsio (approved):

PR #121, opened 2024-05-16
Preload Ollama models before inference and unload afterwards

bauersimon:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -177,6 +177,7 @@ func (command *Evaluate) Execute(args []string) (err error) {
 
 	// Gather models.
 	modelsSelected := map[string]model.Model{}
+	providerForModel := map[model.Model]provider.Provider{}
```
Munsio, line 180: I would do a function on the Model to get its associated provider as all models need to have a provider in the first place. It is a required Parent -> Child dependency right?

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -414,3 +420,25 @@ func writeCSVs(resultPath string, assessments report.AssessmentPerModelPerLangua
 
 	return nil
 }
+
+// withLoadedModel loads the model for the duration of the given function if supported by the model's provider.
+func withLoadedModel(log *log.Logger, model model.Model, providerForModel map[model.Model]provider.Provider, function func()) {
```
Munsio, line 425: We do have the `model` already so no need to pass the whole map, would only pass the single Provider.

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: *sigh* provider already imports model and it would result in a cyclic dependency. I say there is something wrong with our structure if that is happening.

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: If you are scared about having no entry on that key, it is impossible as the list of models is also built when the list of providers is built. 

Should we encounter an no key error than we are manipulating the list of models and having one without a provider is in itself highly illegal.

Munsio (dismissed):

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: Yeeeahhhh.... I was thinking about that but... I guess we can risk it?

bauersimon:
@Munsio plz have another look... also added new tests (though that will change a bit with the changes from the other PR)

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: @bauersimon Please move that to a Follow-Up

tools/ollama.go
```diff
@@ -337,3 +338,39 @@ func OllamaPull(logger *log.Logger, binaryPath string, url string, modelName str
...
+	if err != nil {
+		panic(err)
+	}
+
+	// Send an empty request with negative keep-alive time to load the model and keep it loaded indefinitely. https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-keep-a-model-loaded-in-memory-or-make-it-unload-immediately
+	response, err := http.Post(ollamaURL, "application/json", bytes.NewBufferString(fmt.Sprintf("{\"model\":%q,\"keep_alive\":-1}", modelName)))
+	if err != nil {
+		return pkgerrors.WithStack(pkgerrors.WithMessage(err, fmt.Sprintf("error loading model %q to Ollama server at %q", modelName, ollamaURL)))
+	} else if response.StatusCode != http.StatusOK {
+		return pkgerrors.Errorf("unexpected status code trying to load model %q to Ollama server at %q: %d", modelName, ollamaURL, response.StatusCode)
```
ruiAzevedo19, line 354: `unexpected status code trying to` -> `unexpected status code when trying to`

evaluate/evaluate.go
```diff
@@ -226,3 +232,21 @@ func Evaluate(ctx *Context) (assessments report.AssessmentPerModelPerLanguagePer
 
 	return assessments, totalScore
 }
+
+// withLoadedModel loads the model for the duration of the given function if supported by the model's provider.
+func withLoadedModel(log *log.Logger, model evalmodel.Model, modelProvider provider.Provider, function func()) {
```
ruiAzevedo19, line 237: Wouldn't `task` be a better term instead of `function`?

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -613,4 +622,157 @@ func TestEvaluateExecute(t *testing.T) {
...
+
+					provider.Register(
+						&struct {
+							provider.Provider
+							provider.Loader
+						}{
+							Provider: providerMock,
+							Loader:   loaderMock,
+						},
+					)
```
ruiAzevedo19, line 750: On #137 we moved the tests in `cmd/eval-dev-quality/cmd/evaluate_test.go` that were registering providers to `evaluate/evaluate_test.go` to avoid concurrency problems. I think we should do the same here.

Munsio: Thanks for mentioning it, that one totally slipped on me but is now fixed. Please check again.

ruiAzevedo19:
@Munsio LGTM

PR #132, opened 2024-05-18
Clean up query attempt code

zimmski:


zimmski:
@ruiAzevedo19 merging this right away. so FYI.

PR #133, opened 2024-05-18
Make sure to use uint64 consistently for metrics and scoring, and allow more task cases by always working on a clean repository

zimmski:


zimmski:
@ruiAzevedo19  Merging directly, just FYI

PR #129, opened 2024-05-17
Do not cancel successive runs if previous runs had problems

bauersimon:

PR #134, opened 2024-05-18
More Java task cases for test generation

zimmski:


zimmski:
@bauersimon this is basically an FYI that i early merge this. Will add even more cases, but using these cases now to change your #129 PR to not require changing repositories.

Issue #131, opened 2024-05-18
Follow-up: Allow to retry a model when it errors 


zimmski:
- [x] https://github.com/symflower/eval-dev-quality/pull/125#discussion_r1605747740
- [x] https://github.com/symflower/eval-dev-quality/pull/125#discussion_r1605747874
- [ ] https://github.com/symflower/eval-dev-quality/pull/125#discussion_r1605748485

@ruiAzevedo19 please take care of this before moving to the next task.

PR #137, opened 2024-05-20
Remove the need to change the provider registery in tests to make test code concurrency safe

ruiAzevedo19:
Part of #131


zimmski:
Taking this PR.

PR #146, opened 2024-05-28
Use empty Git config in temporary repositories

### LINKED ISSUE #145: The git repository change requires the GPG password
Currently the workflow is using the systems git configuration which in our case does require a GPG signing of the commits which is not necessary and can be annoying with password inputs.

### Tasks:
- [x] Check how to use a git default configuration with those "imaginary" git repos

### Notes:
- https://gist.github.com/webarchitect609/7eb524ef9eff00779e3b3ab0a5005923

bauersimon:
Closes #145

### DIFF
--- a/evaluate/repository.go
+++ b/evaluate/repository.go
@@ -9,6 +9,7 @@ import (
 
 	pkgerrors "github.com/pkg/errors"
 	"github.com/zimmski/osutil"
+	"github.com/zimmski/osutil/bytesutil"
 
 	"github.com/symflower/eval-dev-quality/evaluate/metrics"
 	"github.com/symflower/eval-dev-quality/language"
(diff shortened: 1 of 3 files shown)


evaluate/repository.go
```diff
@@ -86,6 +87,26 @@ func TemporaryRepository(logger *log.Logger, dataPath string) (temporaryReposito
 		return "", cleanup, pkgerrors.WithStack(err)
 	}
 
+	// Add a default git configuration.
+	if err := os.MkdirAll(filepath.Join(temporaryRepositoryPath, ".git"), 0700); err != nil {
+		return "", cleanup, pkgerrors.WithStack(err)
+	}
+	if err := os.WriteFile(filepath.Join(temporaryRepositoryPath, ".git", "config"), bytesutil.TrimIndentations([]byte(`
```
Munsio, line 94: Instead of writing a file I would simply pass `GIT_CONFIG_NOSYSTEM` https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGNOSYSTEM as environement variable to all `CommandWithResult` calls and then set the `dummy-name` stuff with `git config user.name "Your Name Here"`

Munsio: Or at least run `git init` with `GIT_CONFIG_NOSYSTEM` and then simply override the `.git/config` file with the dummy values.

Munsio (changes requested):

evaluate/repository.go
bauersimon: But I think even `GIT_CONFIG_NOSYSTEM` only disables the system config but then the global user config would still be read, right? And changing only the username does not i.e. reset the GPG signing to the default disabled state then.

So we could run `git init` and then `git config user.name ...` and `git config user.mail ...` though that is two subprocess spawns extra vs. one file operation, that we run for each repository setup?

Munsio: its not worth it if it is working.

Munsio (approved):

PR #148, opened 2024-05-29
Reset repository per task

bauersimon:


evaluate/evaluate_test.go
```diff
@@ -318,19 +318,16 @@ func TestEvaluate(t *testing.T) {
...
-			require.NoError(t, os.WriteFile(filepath.Join(args.String(2), "plain_test.go"), []byte("package plain\nimport \"testing\"\nfunc TestFunction(t *testing.T){}"), 0600))
-		}
 		generateTestsForFilePlainSuccessMetrics := metrics.Assessments{
 			metrics.AssessmentKeyProcessingTime: 1,
 		}
 		generateTestsForFilePlainError := errors.New("generateTestsForFile error")
 
 		generateSuccess := func(mockedModel *modeltesting.MockModel) {
-			mockedModel.On("GenerateTestsForFile", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(generateTestsForFilePlainSuccessMetrics, nil).Run(generateTestsForFilePlainSuccess).Once()
+			mockedModel.RegisterGenerateSuccess(t, "plain_test.go", "package plain\nimport \"testing\"\nfunc TestFunction(t *testing.T){}", generateTestsForFilePlainSuccessMetrics).Once()
```
Munsio: The plain test file name and content are always the same. Can we extract those into a struct inside this test file?

```go
testFile struct {
  Path string
  Content string
}

testFiles := map[string]testFile
```

`RegisterGenerateSuccess(t, testFiles["plain"].Path, testFiles["plain"].Content, `

I guess it will otherwise be copied more and more as soon as we are adding additional tests for additional functionality.

evaluate/repository.go
```diff
@@ -38,6 +38,10 @@ func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, la
 
 	repositoryAssessment = metrics.NewAssessments()
 	for _, filePath := range filePaths {
+		if err := ResetTemporaryRepository(logger, testDataPath); err != nil {
+			logger.Panicf("ERROR: unable to reset temporary repository path: %s", err)
```
Munsio, line 42: Should we add the error to problems and continue here or return an error as we are unable to process the rest of files if there is an error while resetting the directory?

Munsio (changes requested):

evaluate/repository.go
bauersimon: I think the repository stuff is quite fundamental to us. If that does not work anymore, there is no point in keeping on going, right?

I think if we return an error, we will automatically keep going with the next model and just accumulate more errors :thinking:

Munsio: Yeah if we are unable to reset all subsequent runs are in shambles. Actually this would be a `panic` imho.

Munsio (approved):

PR #161, opened 2024-06-05
Do not ignore coverage count if there are failing tests

ahumenberger:


tools/symflower.go
```diff
@@ -91,7 +91,7 @@ func (*symflower) CheckVersion(logger *log.Logger, binaryPath string) (err error
 }
 
 // SymflowerVersionRequired holds the version of Symflower required for this revision of the evaluation.
-const SymflowerVersionRequired = "36800"
+const SymflowerVersionRequired = "37153"
```
bauersimon, line 94: Would be nice to know what fix... cause otherwise the reason is quite generic... sure we are using a new version because it is better than the old one but that is always the case.

language/golang/language.go
```diff
@@ -106,8 +104,33 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
...
+		testSummary := languageGoTestsErrorMatch.FindStringSubmatch(commandOutput)
+		if len(testSummary) > 0 {
+			testCount, e := strconv.ParseInt(testSummary[1], 10, 64)
+			if e != nil {
+				return 0, problems, pkgerrors.WithStack(e)
+			}
+			errorCount, e := strconv.ParseInt(testSummary[2], 10, 64)
+			if e != nil {
+				return 0, problems, pkgerrors.WithStack(e)
+			}
```
bauersimon, line 120: Could use `strconv.Atoi`.

language/golang/language.go
```diff
@@ -106,8 +104,33 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
...
+		}
+
+		problems = append(problems, pkgerrors.WithMessage(pkgerrors.WithStack(err), commandOutput))
+	}
+
+	coverage, err = language.CoverageObjectCountOfFile(coverageFilePath)
+	if err != nil {
+		problems = append(problems, pkgerrors.WithMessage(pkgerrors.WithStack(err), commandOutput))
 	}
```
bauersimon, line 133: I think if the coverage counting failed, it should still be a hard error, no?

language/golang/language.go
```diff
@@ -106,8 +104,33 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
 		Directory: repositoryPath,
 	})
 	if err != nil {
-		return 0, pkgerrors.WithMessage(pkgerrors.WithStack(err), commandOutput)
+		if strings.Contains(commandOutput, "ERROR: no test files found") {
+			return 0, problems, pkgerrors.WithMessage(pkgerrors.WithStack(err), commandOutput)
+		}
+
+		testSummary := languageGoTestsErrorMatch.FindStringSubmatch(commandOutput)
+		if len(testSummary) > 0 {
```
bauersimon, line 112: I think if we cannot find the test summary it should be a hard error.

language/golang/language_test.go
```diff
@@ -132,6 +139,45 @@ func TestLanguageExecute(t *testing.T) {
...
+						}{
+								{"i equal 1", args{1}, 0},
+								{"i equal 2", args{2}, 0},
+						}
+						for _, tt := range tests {
+								if got := simpleIfElse(tt.args.i); got != tt.want {
+										t.Errorf("%q. simpleIfElse(%v) = %v, want %v", tt.name, tt.args.i, got, tt.want)
+								}
+						}
+					}
```
bauersimon, line 171: Is this an actual LLM response? Took me quite a while to figure out what  is going on. Would it be possible to just call our `plain` function (or even this simple-if-else stuff) and then do `assert.True(t, false)` in the very end?

bauersimon: Also for the new Java test case below...

bauersimon (changes requested):

tools/symflower.go
ahumenberger: The commit message state "The fix makes sure that coverage is still reported even if there are failing tests." Not enough?

language/golang/language.go
ahumenberger: Yes. Forgot to change that back in the final version.

language/golang/language.go
ahumenberger: No. We are explicitly looking for "DONE x tests, x error". If there are no errors, then it's "DONE x tests" only. And we are only interested in the case where there are errors.

language/golang/language_test.go
ahumenberger: So you mean using valid test cases, but just make the test fail with a `assert.True(t, false)`. Would probably work, but I don't think this is a blocker.

ahumenberger: I'd rather add a comment to the wrong test case stating this is actual an invalid test case.

tools/symflower.go
bauersimon: ah in the message body... sry... yeah...

language/golang/language.go
ahumenberger: Also, we are just looking in the error case. And in the error case we cannot rely on the summary being there.

language/golang/language_test.go
bauersimon: I meant just using the simplest possible test file to trigger the behavior we need (I.e. produce some coverage and fail). Instead of an actual LLM test file that is super sophisticated with table driven stuff etc..

But no blocker, yes.

zimmski: Yep please reduce.

- One Test has a assert.True(t, true) -> passes
- One Test has a asesrt.True(t, false) -> fails

done no need for logic, no need for an implementation. this is just about `symflower test` being executed.

ahumenberger: > I meant just using the simplest possible test file to trigger the behavior we need

That's what it is basically. I want source code with at least two branches. And then you can have a test file which is table-driven or not. I decided to go table-driven since this is the style from the bug report.

ahumenberger: added the comments as suggested

language/golang/language.go
bauersimon: > If there are no errors, then it's "DONE x tests" only.

So then there are no errors, but... we have `err != nil`. Sound like a hard error to me.

language/golang/language_test.go
ahumenberger: > this is just about symflower test being executed.

I don't agree. This is about getting the correct coverage even if a test fails. So we need an implementation.

language/golang/language.go
ahumenberger: If there is a failing test then it is "DONE 1 tests, 1 failure", but we still have `err != nil` because the exist status is `1`.

bauersimon: Right. And you said if our regex does not find anything, we might have "Done 1 tests" which means no error? But we are still in the block `err != nil` so... something must be off? So hard error?

ahumenberger: I think we went straight into confusion. Let me try again.

We have these scenarios:

- DONE 1 tests
- DONE 0 tests, 1 error -> exit status 1
- DONE 1 tests, 1 failure -> exit status 1

That is, in the first case we have `err == nil`, so we don't need the regex.
For the other two cases we have `err != nil`, but only the first one is a hard error since this indicates a syntax error. For the last case it's just a soft error since we still can get a coverage.

ahumenberger: We can turn the logic to just look for "DONE 1 tests, 1 failure", and everything else is a hard error.

ahumenberger: > We can turn the logic to just look for "DONE 1 tests, 1 failure", and everything else is a hard error.

did that.

language/golang/language_test.go
bauersimon: Will try to reduce the test value.

bauersimon (approved):

PR #162, opened 2024-06-05
Check for Java and Go compilation errors when building a project, for further compile error code repairing task

ruiAzevedo19:


language/golang/mistakes.go
```diff
@@ -0,0 +1,44 @@
...
+		},
+
+		Directory: repositoryPath,
+	})
+	if err != nil {
+		if output != "" {
+			return extractMistakes(output), nil
+		}
+
+		return nil, pkgerrors.WithStack(err)
```
ahumenberger, line 27: `return nil, pkgerrors.Wrap(err, "no output to extract errors from")`

language/golang/mistakes.go
```diff
@@ -0,0 +1,44 @@
...
+var mistakesRe = regexp.MustCompile(`(?m)^.*:\d+:\d+:.*$`)
+
+func extractMistakes(rawMistakes string) (mistakes []string) {
+	results := mistakesRe.FindAllStringSubmatch(rawMistakes, -1)
+
+	for _, result := range results {
+		mistakes = append(mistakes, result[0])
+	}
+
+	return mistakes
```
ahumenberger, line 43: Since you are just accessing the first element of the result, the following should suffice: `return mistakesRe.FindAllString(rawMistakes, -1)`

language/golang/mistakes_test.go
```diff
@@ -0,0 +1,97 @@
...
+		})
+	}
+
+	validate(t, &testCase{
+		Name: "Function without opening bracket",
+
+		RepositoryPath: filepath.Join("..", "..", "testdata", "golang", "mistakes"),
+
+		ExpectedMistakes: []string{
+			"." + string(os.PathSeparator) + "functionWithoutOpeningBracket.go" + ":3:1: syntax error: non-declaration statement outside function body",
```
ahumenberger, line 49: `filepath.Join(".","functionWithoutOpeningBracket.go")`

language/golang/mistakes_test.go
```diff
@@ -0,0 +1,97 @@
...
+			"./foobar.go:4:2: syntax error: non-declaration statement outside function body",
+		},
+	})
+	validate(t, &testCase{
+		Name: "Multiple mistakes",
+
+		RawMistakes: `foobar
+# foobar
+./foobar.go:3:1: expected 'IDENT', found 'func'
+./foobar.go:4:2: syntax error: non-declaration statement outside function body
```
ahumenberger, line 88: Please indent and use `bytesutil.StringTrimIndentations` for these tests.

language/java/mistakes.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+	if err != nil {
+		if output != "" {
+			return extractMistakes(output), nil
+		}
+
+		return nil, pkgerrors.WithStack(err)
+	}
+
+	return nil, nil
```
ahumenberger, line 34: Same suggestions as I did for the Go part.

language/java/mistakes.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// mistakesRe defines the structure of a Java mistake.
+var mistakesRe = regexp.MustCompile(`(?m)\[ERROR\] (.*:\[\d+,\d+\].*)$`)
+
+func extractMistakes(rawMistakes string) (mistakes []string) {
+	m := map[string]bool{}
```
ahumenberger, line 41: Why a map? Can there be multiple entries with the same error message?

language/java/mistakes_test.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+func TestMistakes(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		Logger         *log.Logger
+		RepositoryPath string
+
+		AssertMistakes func(repositoryPath string, mistakes []string)
```
ahumenberger, line 20: Why do we have `AssertMistakes` here and `ExpectedMistakes` for Go?

language/java/mistakes_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	validate(t, &testCase{
+		Name: "Method without opening brackets",
+
+		RepositoryPath: filepath.Join("..", "..", "testdata", "java", "mistakes"),
+
+		AssertMistakes: func(repositoryPath string, mistakes []string) {
+			assert.Contains(t, mistakes[0], "src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected")
+			assert.Contains(t, mistakes[1], "src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interface, or enum expected")
```
ahumenberger, line 50: Why contains, and not an exact match?

ahumenberger (changes requested):

language/golang/mistakes_test.go
ruiAzevedo19: That does not work, `filepath.Join(".","functionWithoutOpeningBracket.go") == functionWithoutOpeningBracket.go` and not `./functionWithoutOpeningBracket.go`

language/java/mistakes.go
ruiAzevedo19: You can check the log in `TestExtractMistakes`,  the same errors are reported after the `COMPILATION ERROR` and after `BUILD FAILURE` logs

language/golang/mistakes.go
```diff
@@ -0,0 +1,44 @@
...
+	})
+	if err != nil {
+		if output != "" {
+			return extractMistakes(output), nil
+		}
+
+		return nil, pkgerrors.WithStack(err)
+	}
+
+	return nil, nil
```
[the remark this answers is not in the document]
ruiAzevedo19: Hum, that's a good question. For me it's a little bit odd returning an error if there are no compiling errors in a project? 

What are your thoughts?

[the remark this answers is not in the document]
zimmski: The task itself should throw an error for the testdata cases, but not this code, to me that is the generic function to receive the info.

language/java/mistakes_test.go
ruiAzevedo19: At first I tried the code below which works fine locally:

```go
expectedMistakes := []string{
     repositoryPath + "/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected",
     repositoryPath + "/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interface, or enum expected",
}

assert.Equal(t, expectedMistakes, mistakes)
```

In the CI:

**Ubuntu**
Works fine

**Macos**
For some reason there is an extra folder in the actual repository path (see logs below)

<details>
 <summary>Log</summary>

```
expected: 
[]string{"/var/folders/lr/439_fwvd3m76p9vy50d57kcc0000gn/T/TestMistakesMethod_without_opening_brackets3653017912/001/mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected", "/var/folders/lr/439_fwvd3m76p9vy50d57kcc0000gn/T/TestMistakesMethod_without_opening_brackets3653017912/001/mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interface, or enum expected"}
actual : 
[]string{"/private/var/folders/lr/439_fwvd3m76p9vy50d57kcc0000gn/T/TestMistakesMethod_without_opening_brackets3653017912/001/mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected", "/private/var/folders/lr/439_fwvd3m76p9vy50d57kcc0000gn/T/TestMistakesMethod_without_opening_brackets3653017912/001/mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interface, or enum expected"}
```

</details>

**Windows**
The actual result is not using the windows back slashes struct for file paths but rather the forward slashes, so we would need to convert the repository file path structure for windows tests (see logs below) 

<details>
 <summary>Log</summary>

```
expected: 
[]string{"C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\TestMistakesMethod_without_opening_brackets3351064100\\001\\mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected", "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\TestMistakesMethod_without_opening_brackets3351064100\\001\\mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interface, or enum expected"}
actual : []string{"/C:/Users/RUNNER~1/AppData/Local/Temp/TestMistakesMethod_without_opening_brackets3351064100/001/mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected\r", "/C:/Users/RUNNER~1/AppData/Local/Temp/TestMistakesMethod_without_opening_brackets3351064100/001/mistakes/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, 
```

</details>

ahumenberger:
Is there a specific reason why we use the wording "mistakes"? To me it looks like more about "compilation errors".

language/golang/mistakes_test.go
ahumenberger: oh right because the result is cleaned.

ruiAzevedo19:
> Is there a specific reason why we use the wording "mistakes"? To me it looks like more about "compilation errors".

@ahumenberger Markus suggested the name "mistakes" for the testdata repository, so I used that name for consistency.

language/java/mistakes.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"golang.org/x/exp/maps"
+
+	pkgerrors "github.com/pkg/errors"
+	"github.com/symflower/eval-dev-quality/log"
+	"github.com/symflower/eval-dev-quality/util"
+)
+
+// Mistakes builds a Java repository and returns the list of mistakes found.
+func Mistakes(logger *log.Logger, repositoryPath string) (mistakes []string, err error) {
```
ruiAzevedo19, line 16: @ahumenberger Maybe we should make this function part of the `language/language.go` interface. 

The reason is that we can simply call `ctx.Language.Mistakes()`  when defining the code repair task. Otherwise we would need to check the language first and call the right `Mistakes` function for each language.

zimmski: Yes. It is language dependent.

language/java/language_test.go
```diff
@@ -249,3 +249,119 @@ func TestLanguageExecute(t *testing.T) {
...
+			[ERROR]
+			[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
+			[ERROR] Re-run Maven using the -X switch to enable full debug logging.
+			[ERROR]
+			[ERROR] For more information about the errors and possible solutions, please read the following articles:
+			[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
+		`),
+		ExpectedMistakes: []string{
+			"/foobar/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected",
+			"/foobar/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interfaceenum, or record expected",
```
bauersimon, line 364: We can remove the `foobar` prefix. It's just a normal path... does not matter if there is a prefix or not.

language/java/language_test.go
```diff
@@ -249,3 +249,119 @@ func TestLanguageExecute(t *testing.T) {
...
+
+func TestMistakes(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		Language *Language
+
+		RepositoryPath string
+
+		AssertMistakes func(repositoryPath string, mistakes []string)
```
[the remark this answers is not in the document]
bauersimon: If the temporary path is a problem, you can go through the actual mistakes in the `validate` and `strings.TrimPrefix(actualMistakes[i], temporaryPath)`.

language/java/language.go
```diff
@@ -112,3 +114,45 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
...
+		}
+
+		return nil, pkgerrors.Wrap(err, "no output to extract errors from")
+	}
+
+	return nil, nil
+}
+
+// mistakesRe defines the structure of a Java mistake.
+var mistakesRe = regexp.MustCompile(`(?m)\[ERROR\] (.*:\[\d+,\d+\].*)$`)
```
bauersimon, line 141: would maybe also assert the `.java:` right before the colon

language/golang/language.go
```diff
@@ -124,3 +124,31 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
...
+			return extractMistakes(output), nil
+		}
+
+		return nil, pkgerrors.Wrap(err, "no output to extract errors from")
+	}
+
+	return nil, nil
+}
+
+// mistakesRe defines the structure of a Go mistake.
```
bauersimon, line 149: "Go compiler error"

language/golang/language.go
```diff
@@ -124,3 +124,31 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
...
+var mistakesRe = regexp.MustCompile(`(?m)^.*:\d+:\d+:.*$`)
```
bauersimon, line 150: assert also `.go:`

language/golang/language_test.go
```diff
@@ -178,3 +178,97 @@ func TestLanguageExecute(t *testing.T) {
 		})
 	})
 }
+
+func TestMistakes(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		Language *Language
```
bauersimon, line 186: Remove language.

language/golang/language_test.go
```diff
@@ -178,3 +178,97 @@ func TestLanguageExecute(t *testing.T) {
...
+		})
+	}
+
+	validate(t, &testCase{
+		Name: "Function without opening bracket",
+
+		RepositoryPath: filepath.Join("..", "..", "testdata", "golang", "mistakes"),
+
+		ExpectedMistakes: []string{
+			"." + string(os.PathSeparator) + "functionWithoutOpeningBracket.go" + ":4:6: syntax error: non-declaration statement outside function body",
```
bauersimon, line 222: Maybe  `filepath.Join(".", "functionWithoutOpeningBracket.go")+":4...`

bauersimon: To clarify it is about the file.

bauersimon (changes requested):

language/java/language.go
```diff
@@ -112,3 +114,45 @@ func (l *Language) Execute(logger *log.Logger, repositoryPath string) (coverage
...
 	return coverage, nil, nil
 }
+
+// Mistakes builds a Java repository and returns the list of mistakes found.
+func (l *Language) Mistakes(logger *log.Logger, repositoryPath string) (mistakes []string, err error) {
+	output, err := util.CommandWithResult(context.Background(), logger, &util.Command{
+		Command: []string{
+			"mvn",
+			"clean",
+			"install",
```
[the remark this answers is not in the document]
ruiAzevedo19: Both solutions don't log all the errors. I think it's better to change this to `mvn compile`

language/language.go
```diff
@@ -24,6 +24,8 @@ type Language interface {
 
 	// Execute invokes the language specific testing on the given repository.
 	Execute(logger *log.Logger, repositoryPath string) (coverage uint64, problems []error, err error)
+	// Mistakes builds a repository and returns the list of mistakes found.
+	Mistakes(logger *log.Logger, repositoryPath string) (mistakes []string, err error)
```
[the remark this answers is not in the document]
ruiAzevedo19: The other commit defines this for Go. I will add an empty implementation for Go in this commit

language/golang/language_test.go
ruiAzevedo19: See https://github.com/symflower/eval-dev-quality/pull/162#discussion_r1629091363

language/java/language_test.go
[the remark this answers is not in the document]
ruiAzevedo19: The trim doesn't work well because AFAIK in Windows the temporary path will have back slashes but by looking into the CI logs the function returns the path with forward slashes, so the trim is not taking place.

Also, for some reason, on MacOs, the function is returning paths in the form of `/private/<temporary-path>`.

I did it in a slightly different way, but still using `ExpectedMistakes []string`

language/java/language.go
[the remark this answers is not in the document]
bauersimon: I think we internally use `mvn compile` as well for `symflower`. But it should log everything :thinking: Have you tried the `-e` flag which shows errors? And the `-fae` flag which "fails at end"?

language/language.go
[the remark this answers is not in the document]
bauersimon: yeah I realized that as well ^^ not crazy terrible, was just confused slightly

language/java/language.go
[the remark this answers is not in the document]
bauersimon: I think the compile also creates `.class` files but that is okay. The caller must take care of cleanup just as for the other cases.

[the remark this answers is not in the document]
ruiAzevedo19: This problem happened when I ran `mvn compile` in a repository with several files with errors (even with the `-e` and `-fae` flags).  For the testdata in this PR is working fine and we are getting all the errors.

The goal for the upcoming issues regarding tasks is to place each file with errors in its own repository, so maybe `mvn compile` is enough.

language/java/language_test.go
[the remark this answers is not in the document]
bauersimon: so maven returns the errors on windows with file path forward slashes? but whyyyyyyy

language/java/language_test.go
```diff
@@ -249,3 +250,121 @@ func TestLanguageExecute(t *testing.T) {
...
+			}()
+
+			java := &Language{}
+			actualMistakes, actualErr := java.Mistakes(logger, repositoryPath)
+			require.NoError(t, actualErr)
+
+			for i, mistake := range actualMistakes {
+				index := strings.Index(mistake, "/src/main/java")
+				require.Greater(t, index, -1)
+				actualMistakes[i] = mistake[index:]
```
bauersimon, line 283: Could you please add a comment that Maven prints paths with slashes even on Windows, so to split off the temporary path we need to be creative?

bauersimon (changes requested):

language/java/language_test.go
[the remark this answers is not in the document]
ruiAzevedo19: Here's the failed CI job where you can see the windows paths: https://github.com/symflower/eval-dev-quality/actions/runs/9466371364/job/26077902331

bauersimon (dismissed):

Issue #165, opened 2024-06-06
Support multiple evaluation tasks


ahumenberger:
As of now, we only have a single evaluation task, that is, "provide tests for this code". In a future release we want to run other tasks as well, like "repair this compilation error". For this, we need structural changes.

**Tasks:**
- [x] Introduce string enum `TaskIdentifier`
- [x] Change `Model` interface such that it has a `RunTask(TaskIdentifier,...)` method
  - [x] Remove `GenerateTestsForFile` from `Model` interface as this becomes a task
- [x] Add a method `IsTaskSupported(TaskIdentifier)` to the `Model` interface to let models specify if they support a specific task.
- [x] Introduce proper `Repository` type
  - [x] Each repository can allow/disallow certain tasks
    - [x] This should not be hardcoded. Therefore a `repository.json` file can be created in the repository's root directory. If no such file exists, all tasks are allowed.
- [x] Create `Task` interface and suitable implementations

PR #166, opened 2024-06-06
Introduce the concept of "tasks" to prepare for different evaluation tasks like "write tests" and "repair code"

ahumenberger:
Part of #165


task/task.go
```diff
@@ -0,0 +1,42 @@
...
+func registerIdentifier(name string) (identifier Identifier) {
+	identifier = Identifier(name)
+	AllIdentifiers = append(AllIdentifiers, identifier)
+
+	return identifier
+}
+
+var (
+	// IdentifierWriteTests holds the identifier for the "write test" task.
+	IdentifierWriteTests = registerIdentifier("write-tests")
```
ahumenberger, line 31: Not sure about the naming. I think before it was called "generate tests", I just noticed that in the end.

ahumenberger:
I think it worth doing https://github.com/symflower/eval-dev-quality/issues/169 first.

task/task.go
zimmski: write tests is fine with me. Since the POV is that we are benchmarking a developer.

Just to note this: in the LLM world we usually talk about CodeGen

task/task.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/symflower/eval-dev-quality/language"
+)
+
+var (
+	// ErrTaskUnsupported indicates that a task is unsupported.
+	ErrTaskUnsupported = errors.New("task unsupported")
+)
+
+// Identifier holds the identifier of a task.
+type Identifier string
```
zimmski, line 16: I like this to be reworded. In evaluate/metrics/assessment.go we have "Key". @bauersimon @ahumenberger any other suggestions? "Identifier" doesn't sound right to me.

evaluate/evaluate_test.go
```diff
@@ -27,6 +27,7 @@ import (
 	modeltesting "github.com/symflower/eval-dev-quality/model/testing"
 	"github.com/symflower/eval-dev-quality/provider"
 	providertesting "github.com/symflower/eval-dev-quality/provider/testing"
+	"github.com/symflower/eval-dev-quality/task"
```
zimmski, line 30: Please update the README as well with these changes. We now have a generic task interface.

model/symflower/symflower.go
```diff
@@ -33,27 +32,47 @@ func (m *Model) ID() (id string) {
 	return "symflower" + provider.ProviderModelSeparator + "symbolic-execution"
 }
 
-// GenerateTestsForFile generates test files for the given implementation file in a repository.
-func (m *Model) GenerateTestsForFile(logger *log.Logger, language language.Language, repositoryPath string, filePath string) (assessment metrics.Assessments, err error) {
```
zimmski, line 37: Please keep that exported function. Fine with converting it to have context parameter, but unexporting it makes using it impossible in other code. And i would like to rework the interface when i am done with the report i need to write. That would mean reverting that as well.

model/model.go
```diff
@@ -2,17 +2,18 @@ package model
...
+	"github.com/symflower/eval-dev-quality/task"
 )
 
 // Model defines a model that can be queried for generations.
 type Model interface {
 	// ID returns the unique ID of this model.
 	ID() (id string)
 
-	// GenerateTestsForFile generates test files for the given implementation file in a repository.
-	GenerateTestsForFile(logger *log.Logger, language language.Language, repositoryPath string, filePath string) (assessments metrics.Assessments, err error)
```
zimmski, line 15: Same as with https://github.com/symflower/eval-dev-quality/pull/166/commits/4e6d0619eb159f7212ab4a713f45d0b404b92271#r1632828265

evaluate/report/csv_test.go
```diff
@@ -32,57 +33,63 @@ func TestGenerateCSVForAssessmentPerModelPerLanguagePerRepository(t *testing.T)
...
-					"some-repository": metrics.Assessments{
-						metrics.AssessmentKeyGenerateTestsForFileCharacterCount: 50,
-						metrics.AssessmentKeyResponseCharacterCount:             100,
-						metrics.AssessmentKeyCoverage:                           1,
-						metrics.AssessmentKeyFilesExecuted:                      2,
-						metrics.AssessmentKeyResponseNoError:                    3,
-						metrics.AssessmentKeyResponseNoExcess:                   4,
-						metrics.AssessmentKeyResponseWithCode:                   5,
-						metrics.AssessmentKeyProcessingTime:                     200,
+					"some-repository": {
```
zimmski, line 57: Please use the type for composite literals.

task/task.go
bauersimon: Each Model, Provider, Tool on the other hand does have an `ID`. But that is a slightly different structure.

bauersimon: Or since it is the only thing that currently identifies what a task is, why not call it `Task`?

bauersimon: And then `AllIdentifiers` -> `AllTasks`.

zimmski: I see. That brings me back to my original request i had with Andi: making the tasks use methods of models/languages instead of letting the models reimplement tasks. Then every task has the same pattern as model/language/provider/...

model/symflower/symflower.go
ahumenberger: Can do that, but I don't think that's how we usually work, unless there is a concrete plan how this will look like in the end. Feels like YAGNI.

> That would mean reverting that as well.

Reverting needs to be done anyhow since the change is already done in this PR.

> And i would like to rework the interface 

Please do a planning for this. I also put some thought into this structure.

task/task.go
ahumenberger: The assessment type is basically a database, so "Key" fits perfectly there, but it does not for tasks. A task identifier identifies a task. E.g. a repository can  list tasks which are supposed to be evaluated. Sounds right to me.

ahumenberger: > making the tasks use methods of models/languages instead of letting the models reimplement tasks.

This is meant to be done as a follow-up. This PR only introduces the primitive concept running different tasks. For this, a "task identifier" is sufficient. Later a `Task` type can be introduced. A "task identifier" is still necessary, because e.g. a repository can specify a list of tasks that should be run.

model/llm/llm.go
```diff
@@ -32,7 +32,7 @@ type Model struct {
 }
 
 // NewModel returns an LLM model corresponding to the given identifier which is queried via the given provider.
-func NewModel(provider provider.Query, modelIdentifier string) model.Model {
+func NewModel(provider provider.Query, modelIdentifier string) *Model {
```
bauersimon, line 35: Not a huge thing but why is this necessary?

evaluate/repository.go
```diff
@@ -19,34 +19,42 @@ import (
...
 )
 
-// Repository evaluate a repository with the given model and language.
-func Repository(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language, testDataPath string, repositoryName string) (repositoryAssessment metrics.Assessments, problems []error, err error) {
-	log, logClose, err := log.WithFile(logger, filepath.Join(resultPath, evalmodel.CleanModelNameForFileSystem(model.ID()), language.ID(), repositoryName+".log"))
+// Repository holds data about a repository.
+type Repository struct {
+	// Name holds the name of the repository.
+	Name string
+	// DataPath holds the path to the repository.
```
bauersimon, line 54: Absolute or relative? Just asking because we usually also have this `TestdataPath` that makes all the repository paths relative.

evaluate/repository.go
```diff
@@ -19,34 +19,42 @@ import (
...
 	if err != nil {
 		return nil, problems, pkgerrors.WithStack(err)
 	}
 
 	repositoryAssessment = metrics.NewAssessments()
 	for _, filePath := range filePaths {
-		if err := ResetTemporaryRepository(logger, testDataPath); err != nil {
+		if err := r.Reset(logger); err != nil {
 			logger.Panicf("ERROR: unable to reset temporary repository path: %s", err)
 		}
```
bauersimon, line 100: Usability nitpick: since we anyways panic here (and I think we always panic if the reset mechanism fails) we could also just panic within `Reset`.

evaluate/repository.go
```diff
@@ -19,16 +20,63 @@ import (
...
+func (rc *repositoryConfiguration) validate() (err error) {
+	if len(rc.Tasks) == 0 {
+		return pkgerrors.Errorf("empty list of tasks in configuration")
+	}
+
+	for _, taskIdentifier := range rc.Tasks {
+		if !task.LookupIdentifier[taskIdentifier] {
+			return pkgerrors.Errorf("task identifier %q unknown", taskIdentifier)
+		}
+	}
```
bauersimon, line 38: Plz wrap errors with stacks.

evaluate/evaluate_test.go
```diff
@@ -535,11 +559,12 @@ func TestEvaluate(t *testing.T) {
...
-						languageGolang: map[string]metrics.Assessments{
-							repositoryPlainPath: map[metrics.AssessmentKey]uint64{},
-							repositoryNextPath:  map[metrics.AssessmentKey]uint64{},
+				ExpectedAssessments: map[evalmodel.Model]map[language.Language]map[string]map[task.Identifier]metrics.Assessments{
+					mockedModel: map[language.Language]map[string]map[task.Identifier]metrics.Assessments{
+						languageGolang: map[string]map[task.Identifier]metrics.Assessments{
+							repositoryPlainPath: map[task.Identifier]metrics.Assessments{
+								task.IdentifierWriteTests: map[metrics.AssessmentKey]uint64{},
+							},
 						},
```
bauersimon, line 568: `repositoryNextPath` is gone. Guess does not matter if its empty or gone, but I think we should keep it?

evaluate/evaluate_test.go
```diff
@@ -121,10 +121,12 @@ func TestEvaluate(t *testing.T) {
...
-					for _, a := range rs {
-						if v, ok := a[metrics.AssessmentKeyProcessingTime]; ok {
-							if assert.Greater(t, v, uint64(0)) {
-								delete(a, metrics.AssessmentKeyProcessingTime)
+					for _, ts := range rs {
+						for _, a := range ts {
+							if v, ok := a[metrics.AssessmentKeyProcessingTime]; ok {
+								if assert.Greater(t, v, uint64(0)) {
+									delete(a, metrics.AssessmentKeyProcessingTime)
+								}
```
bauersimon, line 129: Out of context.

We already had a logic that ignores all nondeterministic parts of assessments. So with https://github.com/symflower/eval-dev-quality/commit/b207e284f4a76054dd3e2bf4259f7d2e744bffe7 we now do that twice in two separate locations. Could you please either resolve this or keep it as a follow-up?

evaluate/repository.go
```diff
@@ -19,16 +20,63 @@ import (
...
+
+	return r.repositoryConfiguration.validate()
+}
+
 // Evaluate evaluates a repository with the given model and language.
-func (r *Repository) Evaluate(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language) (repositoryAssessment metrics.Assessments, problems []error, err error) {
+func (r *Repository) Evaluate(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language, taskIdentifier task.Identifier) (repositoryAssessment metrics.Assessments, problems []error, err error) {
 	log, logClose, err := log.WithFile(logger, filepath.Join(resultPath, evalmodel.CleanModelNameForFileSystem(model.ID()), language.ID(), r.Name+".log"))
 	if err != nil {
 		return nil, nil, err
```
bauersimon, line 82: Please  consider changing the logging to also incorporate the task, and add the task also to the log call below (which I cannot select for review - thanks GitHub!).

bauersimon (changes requested):

model/llm/llm.go
ahumenberger: Because in the tests we create a model with `NewModel` and then call `GenerateTestsFor....` and this is not a method of `model.Model` anymore.

evaluate/evaluate_test.go
ahumenberger: Why do you think we should keep it?

The test is the following: "Never solving basic checks leads to exclusion". So this is IMO fine if there is nothing for `repositoryNextPath`.

Also the reason why it was there before is because the nested maps were preinitialized for all models, languages and repositories. We would need to do the same for tasks. But not every model supports every task, and not every repository allows all tasks. So this gets complicated to preinitialize. It's way easier to initialize on demand. So unless there is a good reason why we need the old version, I'd keep  this one.

task/task.go
```diff
@@ -0,0 +1,42 @@
...
+// AllIdentifiers holds all available task identifiers.
+var AllIdentifiers []Identifier
+
+// registerIdentifier registers the given identifier and makes it available.
+func registerIdentifier(name string) (identifier Identifier) {
+	identifier = Identifier(name)
+	AllIdentifiers = append(AllIdentifiers, identifier)
+
+	return identifier
+}
```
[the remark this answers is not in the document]
ahumenberger: In a later commit I'm introducing a map as well. Added a check now to make sure that identifiers are not added twice.

evaluate/repository.go
bauersimon: ahhhh I confused that one.

evaluate/repository.go
ahumenberger: Nah, not a friend of that. It should be the caller's decision what to do with the error.

evaluate/report/csv_test.go
ahumenberger: The tests were already like this. And since I'm planning on changing this with #169 I'm not really eager to change this now.

evaluate/repository.go
ahumenberger: absolute. changed the docstring for the field

evaluate/evaluate_test.go
ahumenberger: Not sure what you mean. If you are talking about the "normalize assessments, then this is the very same location.

model/model.go
ahumenberger: See https://github.com/symflower/eval-dev-quality/pull/166#discussion_r1634653364

task/task.go
bauersimon: Then let's do `Key` for now? And when we touch it again, we can see what makes most sense then.

ahumenberger: Since I think "Identifier" is totally fine, and there are no better suggestions, I'm going to leave it.

evaluate/evaluate_test.go
bauersimon: It is... sorry for not clarifying. There already is `metricstesting.AssertAssessmentsEqual`. And now this as well.

evaluate/repository.go
bauersimon: Should we maybe change the log file location as well? I.e. different files for different tasks, otherwise they will clump up in one log file.

evaluate/evaluate_test.go
bauersimon: its not a big deal... just good to keep an eye out in the future

evaluate/repository.go
ahumenberger: true true true

evaluate/evaluate_test.go
ahumenberger: Briefly looked at this. But I don't want to touch this right now. Might be that some tests are failing again, and this would delay merging the PR.

ahumenberger: Added to #169

bauersimon (approved):

Issue #168, opened 2024-06-07
Evaluation task: Code repair


ruiAzevedo19:
#### Goal

Given source code with compilation errors, the model needs to repair the code such that the source code compiles. The response is validated by executing predefined tests making sure that the implementation itself is not altered.

 #### PRs
  - https://github.com/symflower/eval-dev-quality/pull/170
  - https://github.com/symflower/eval-dev-quality/pull/192
  - https://github.com/symflower/eval-dev-quality/pull/197

#### Follow-up
- https://github.com/symflower/eval-dev-quality/issues/200
#### TODOs  
- [x] Testdata
  - Examples 
    -  function opening brackets are missing
    - type is missing
    - type is wrong
    - import is missing
    - variable is not declared
  - For each case:
    - [x] generate test with `symflower unit-tests`
    - [x] check the tests are passing
    - [x] add a mistake to the implementation
    - [x] commit   
- [x] Implementation
  - [x] Define a new task identifier: `code-repair`
  - [x] For `symflower` model define this task as unsupported because we always generate deterministic tests
  - [x] For LLM models
    - [x] Define the new task as supported 
    - [x] Create an interface for tasks
      - [x] Interface: `Task`
      - [x] Methods
        - [x] `Run(repository) (assessment, err)`: run the task for the given repository and return the assessments
        - [x] `Identifier`: returns the task identifier
    - [x] Define tasks
      - [x] `TaskWriteTests` 
        - [x] The `Run` method is basically what we already have in `evaluate/repository.go:Evaluate`
        - [x] Remove `evaluate/repository.go:Evaluate` since is now part of the task  
      - [x]  `TaskCodeRepair` 
        - [x] The `Run` method is responsible to only run the task for source code files (filter out test files and other files)
          - [x] The method must range over the sub-directories  in `mistakes` testdata and and run the code repair task for each sub-directory
          - [x] Add two methods to the language interface
            - [x]  `DefaultFileExtension` returns the language file extension
            - [x] `DefaultTestFileSuffix` returns the language test file suffix, i.e., `_test.go` for Go and `Test.java` for Java
            - Note: this will be used to easily filter out files
    - [x] Calling the `Run` method
      - [x] replace the call `temporaryRepository.Evaluate(...)` in `evaluate/evaluate.go:Evaluate` with the task `Run` method
[...]

PR #170, opened 2024-06-07
Allow LLMs to do code repair if there is any compilation error when building a repository

ruiAzevedo19:
Part of #168


model/llm/llm.go
```diff
@@ -182,6 +219,59 @@ func (m *Model) queryModel(log *log.Logger, request string) (assessment metrics.
...
 
+// repairSourceCodeFile asks the model to repair the code in case any compilation error occurs in a repository source code.
+func (m *Model) repairSourceCodeFile(ctx task.Context) (assessment metrics.Assessments, err error) {
+	if !strings.HasSuffix(ctx.FilePath, ".go") && !strings.HasSuffix(ctx.FilePath, ".java") {
+		return nil, nil
+	}
+
+	if strings.HasSuffix(ctx.FilePath, "_test.go") || strings.HasSuffix(ctx.FilePath, "Test.java") {
+		return nil, nil
+	}
```
ruiAzevedo19, line 230: This task is called for each file in a repository. Until know, all the repositories only had source files (`.go` or `.java`) but now we also have tests inside a repository (`_test.go` or `Test.java`) so we need to skip running the task for those files. 

Wondering if there is a more elegant way to do this.

evaluate/repository.go
```diff
@@ -139,10 +147,19 @@ func contextForTask(taskIdentifier task.Identifier, language language.Language,
...
 	switch taskIdentifier {
+	case task.IdentifierCodeRepair:
+		mistakes, err := language.Mistakes(logger, repositoryPath)
+		if err != nil {
+			return context, err
+		}
+
+		context.Arguments = mistakes
+
+		return context, nil
```
ruiAzevedo19, line 158: If there are no mistakes we don't need to execute the task at all. Should we also return a boolean indicating if a task should be skipped?

evaluate/repository_test.go
```diff
@@ -116,6 +123,107 @@ func TestRepository(t *testing.T) {
...
+			llm := llm.NewModel(mockProvider, modelId)
+
+			validate(t, &testCase{
+				Name: "Go",
+
+				Model:          llm,
+				Language:       &golang.Language{},
+				TaskIdentifier: task.IdentifierCodeRepair,
+				TestDataPath:   filepath.Join("..", "testdata"),
+				RepositoryPath: filepath.Join("golang", "mistakes", "openingBracketIsMissing"),
```
bauersimon, line 155: these tests will change with the modifications that we discussed

evaluate/repository_test.go
```diff
@@ -314,16 +427,6 @@ func TestRepositoryLoadConfiguration(t *testing.T) {
...
-		MutationBefore: func(t *testing.T, repositoryPath string) {
-			configuration := bytesutil.StringTrimIndentations(`
-				{
-					"tasks": [
-						"write-tests"
-					]
-				}
-			`)
-			assert.NoError(t, os.WriteFile(filepath.Join(repositoryPath, "repository.json"), []byte(configuration), 0600))
-		},
```
bauersimon, line 326: I think this needs to go into the previous commit where you added the config to the plain directory!

evaluate/repository_test.go
```diff
@@ -304,6 +413,10 @@ func TestRepositoryLoadConfiguration(t *testing.T) {
 		TestDataPath:   filepath.Join("..", "testdata"),
 		RepositoryPath: filepath.Join("golang", "plain"),
 
+		MutationBefore: func(t *testing.T, repositoryPath string) {
+			assert.NoError(t, os.Remove(filepath.Join(repositoryPath, "repository.json")))
+		},
+
```
bauersimon, line 419: I think this needs to go into the previous commit where you added the config to the plain directory!

go.mod
```diff
@@ -2,36 +2,10 @@ module github.com/symflower/eval-dev-quality
...
-	github.com/rogpeppe/go-internal v1.9.0 // indirect
-	github.com/schollz/progressbar/v3 v3.14.2 // indirect
-	github.com/stretchr/objx v0.5.2 // indirect
-	github.com/termie/go-shutil v0.0.0-20140729215957-bcacb06fecae // indirect
-	github.com/ulikunitz/xz v0.5.11 // indirect
-	github.com/yuin/goldmark v1.7.0 // indirect
-	golang.org/x/image v0.11.0 // indirect
-	golang.org/x/sys v0.18.0 // indirect
-	golang.org/x/term v0.18.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
```
bauersimon: that looks very wrong.

language/golang/language_test.go
```diff
@@ -212,10 +212,10 @@ func TestMistakes(t *testing.T) {
 	validate(t, &testCase{
 		Name: "Function without opening bracket",
 
-		RepositoryPath: filepath.Join("..", "..", "testdata", "golang", "mistakes"),
+		RepositoryPath: filepath.Join("..", "..", "testdata", "golang", "mistakes", "openingBracketIsMissing"),
 
 		ExpectedMistakes: []string{
-			"." + string(os.PathSeparator) + "functionWithoutOpeningBracket.go" + ":4:6: syntax error: non-declaration statement outside function body",
+			"." + string(os.PathSeparator) + "openingBracketIsMissing.go" + ":4:2: syntax error: non-declaration statement outside function body",
```
bauersimon, line 218: need to adapt this as well cause it was renamed

language/java/language_test.go
```diff
@@ -291,11 +291,14 @@ func TestMistakes(t *testing.T) {
...
+		RepositoryPath: filepath.Join("..", "..", "testdata", "java", "mistakes", "openingBracketIsMissing"),
 
 		ExpectedMistakes: []string{
-			"/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[4,61] ';' expected",
-			"/src/main/java/com/eval/MethodWithoutOpeningBracket.java:[7,1] class, interface, or enum expected",
+			"/src/main/java/com/eval/OpeningBracketIsMissing.java:[12,17] illegal start of type",
+			"/src/main/java/com/eval/OpeningBracketIsMissing.java:[14,1] class, interface, or enum expected",
+			"/src/main/java/com/eval/OpeningBracketIsMissing.java:[4,57] ';' expected",
+			"/src/main/java/com/eval/OpeningBracketIsMissing.java:[8,17] illegal start of type",
+			"/src/main/java/com/eval/OpeningBracketIsMissing.java:[8,25] illegal start of type",
```
bauersimon, line 301: got renamed

model/llm/llm.go
```diff
@@ -145,29 +163,21 @@ func (m *Model) generateTestsForFile(ctx task.Context) (assessment metrics.Asses
...
-			ctx.Logger.Printf("Attempt %d/%d: %s", n+1, m.queryAttempts, err)
+			log.Printf("Attempt %d/%d: %s", n+1, m.queryAttempts, err)
 		}),
 	); err != nil {
-		return nil, err
+		return nil, "", err
 	}
 
-	assessment, testContent, err := prompt.ParseResponse(response)
+	assessment, testContent, err = prompt.ParseResponse(response)
```
bauersimon, line 211: Sorry for noticing now, would extract the parsing logic from the query helper. Cause that parsing might not always be what we want. The query model should really only be the query and return the response.

bauersimon (changes requested):

model/llm/llm.go
bauersimon: And while we are at it, just call the function `query`. Otherwise it is `Model.queryModel` which is doubled.

bauersimon: function not renamed yet :disappointed:

evaluate/repository.go
```diff
@@ -134,6 +127,25 @@ func (r *Repository) Evaluate(logger *log.Logger, resultPath string, model evalm
...
+
+		Logger: logger,
+	}
+
+	switch taskIdentifier {
+	case task.IdentifierWriteTests:
+		return context
+	default:
+		return context
+	}
```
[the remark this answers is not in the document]
ruiAzevedo19: No longer applies since the logic changed.

evaluate/repository.go
```diff
@@ -99,13 +99,21 @@ func (r *Repository) Evaluate(logger *log.Logger, resultPath string, model evalm
...
+		}
 		assessments, err := model.RunTask(ctx, taskIdentifier)
 		if err != nil {
 			problems = append(problems, pkgerrors.WithMessage(err, filePath))
 
 			continue
 		}
+		if assessments == nil && err == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
ruiAzevedo19: No longer applies since the logic changed.

task/task.go
```diff
@@ -1,15 +1,122 @@
...
 type Task interface {
 	// Identifier returns the task identifier.
 	Identifier() taskcontext.Identifier
 
 	// Run runs a task in a given repository.
 	Run(repository *repository.Repository) (assessments metrics.Assessments, problems []error, err error)
 }
+
+// GetTaskByTaskIdentifier returns a task based on the task identifier.
+func GetTaskByTaskIdentifier(taskIdentifier taskcontext.Identifier, logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language) (task Task, err error) {
```
ahumenberger, line 26: `TaskForIdentifier`

task/task.go
```diff
@@ -1,15 +1,122 @@
...
+	Model evalmodel.Model
+
+	// Logger holds the logger for this tasks.
+	Logger *log.Logger
+}
+
+var _ Task = (*TaskWriteTests)(nil)
+
+// NewTaskWriteTests returns a write test task.
+func NewTaskWriteTests(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language) (task Task) {
```
ahumenberger, line 52: `newTaskWriteTest` as the single constructor for tasks should be `TaskForIdentifier`

task/task.go
```diff
@@ -1,15 +1,122 @@
...
+	switch taskIdentifier {
+	case taskcontext.IdentifierWriteTests:
+		return NewTaskWriteTests(logger, resultPath, model, language), nil
+	default:
+		return nil, taskcontext.ErrTaskUnsupported
+	}
+}
+
+// TaskWriteTests holds the write test task.
+type TaskWriteTests struct {
```
ahumenberger, line 36: The task interface and its implementations should be separated IMO. Ideally `Context`, `Identifier` and `Task` itself should be in `github.com/symflower/eval-dev-quality/task`. And the implementations could be e.g. in `github.com/symflower/eval-dev-quality/evaluate/task`.

ahumenberger: Does this resolve some of the cyclic imports you experienced?

model/llm/llm.go
```diff
@@ -184,6 +221,59 @@ func (m *Model) query(log *log.Logger, request string) (response string, duratio
 	return response, duration, nil
 }
 
+// repairSourceCodeFile asks the model to repair the code in case any compilation error occurs in a repository source code.
+func (m *Model) repairSourceCodeFile(ctx taskcontext.Context) (assessment metrics.Assessments, err error) {
+	assessment = map[metrics.AssessmentKey]uint64{}
+
+	mistakes, ok := ctx.Arguments.([]string)
```
ahumenberger, line 228: Please create a

```go
type TaskCodeRepairArguments struct {
  Mistakes []string
}
```

to pass as arguments.

task/task.go
```diff
@@ -120,3 +121,115 @@ func (t *TaskWriteTests) Run(repository *repository.Repository) (repositoryAsses
...
+	Model evalmodel.Model
+
+	// Logger holds the logger for this tasks.
+	Logger *log.Logger
+}
+
+var _ Task = (*TaskCodeRepair)(nil)
+
+// NewCodeRepairTask returns a code repair task.
+func NewCodeRepairTask(logger *log.Logger, resultPath string, model evalmodel.Model, language language.Language) (task Task) {
```
ahumenberger, line 142: `newTaskCodeRepair`

ahumenberger (changes requested):

task/task.go
ahumenberger: @bauersimon Thoughts?

model/llm/llm.go
ahumenberger: Or `TaskArgumentsCodeRepair`, not sure

@bauersimon

task/task.go
ahumenberger: >  Ideally Context, Identifier and Task itself should be in github.com/symflower/eval-dev-quality/task

I suppose only the type `Identifier` should be in the `task` package. `IdentifierWriteTests` should then probably be in `evaluate/task`.

testdata/java/mistakes/openingBracketMissing/src/test/java/com/eval/OpeningBracketMissingTest.java
```diff
@@ -8,7 +8,7 @@ public class OpeningBracketMissingTest {
 	public void openingBracketMissing1() {
 		int x = -1;
 		int expected = -1;
-		int actual = OpeningBracketIsMissing.openingBracketMissing(x);
+		int actual = OpeningBracketMissing.openingBracketMissing(x);
 
 		assertEquals(expected, actual);
 	}
```
bauersimon, line 14: Should we maybe add some tests that all repositories have to build?

task/task.go
bauersimon: The cyclic imports should be gone as long as the `Identifier` and `Context` are in a separate package, cause both the task logic and the model need them, so they cannot be in the same package than the task logic.

Hmmmm. Splitting interface and implementation to completely different packages seems a bit weird. And honestly I think a `task` is such an essential part of `evaluate` that I would move everything there. And then we would have `evaluate/task` and `evaluate/task/context`.

task/task.go
```diff
@@ -1,15 +1,122 @@
...
+func (t *TaskWriteTests) Run(repository *repository.Repository) (repositoryAssessment metrics.Assessments, problems []error, err error) {
+	log, logClose, err := log.WithFile(t.Logger, filepath.Join(t.ResultPath, string(t.Identifier()), evalmodel.CleanModelNameForFileSystem(t.Model.ID()), t.Language.ID(), repository.Name+".log"))
+	if err != nil {
+		return nil, nil, err
+	}
+	defer logClose()
+
+	log.Printf("Evaluating model %q on task %q using language %q and repository %q", t.Model.ID(), t.Identifier(), t.Language.ID(), repository.Name)
+	defer func() {
+		log.Printf("Evaluated model %q on task %q using language %q and repository %q: encountered %d problems: %+v", t.Model.ID(), t.Identifier(), t.Language.ID(), repository.Name, len(problems), problems)
```
bauersimon, line 76: All the logging business will be the same for all tasks, so should probably create a helper and/or move into `evaluate/evaluate.go`.

evaluate/repository/repository_test.go
```diff
@@ -194,7 +53,7 @@ func TestTemporaryRepository(t *testing.T) {
 	validate(t, &testCase{
 		Name: "Create temporary path with git repository",
 
-		TestDataPath:   filepath.Join("..", "testdata"),
+		TestDataPath:   filepath.Join("..", "..", "testdata"),
```
bauersimon, line 56: These are changes from a previous commit where the repository was moved.

language/golang/language.go
```diff
@@ -75,6 +75,16 @@ func (l *Language) TestFramework() (testFramework string) {
...
+// DefaultFileExtension returns the default file extension.
+func (l *Language) DefaultFileExtension() string {
+	return ".go"
+}
+
+// DefaultTestFileSuffix returns the default test file suffix.
+func (l *Language) DefaultTestFileSuffix() string {
+	return "_test.go"
+}
+
```
bauersimon, line 87: always move such easy refactors to the front of the PR for early merging.

language/java/language.go
```diff
@@ -88,6 +88,16 @@ func (l *Language) TestFramework() (testFramework string) {
...
+// DefaultFileExtension returns the default file extension.
+func (l *Language) DefaultFileExtension() string {
+	return ".java"
+}
+
+// DefaultTestFileSuffix returns the default test file suffix.
+func (l *Language) DefaultTestFileSuffix() string {
+	return "Test.java"
+}
+
```
bauersimon, line 100: nit: for java we sometimes also have `Tests.java` so since this is just for filtering we could also say `IsTestFilePath` and use a regex that allows for both. But since we decide how the examples look, it's fine!

model/llm/llm.go
bauersimon: second one sounds noice!

model/llm/llm.go
```diff
@@ -184,6 +221,59 @@ func (m *Model) query(log *log.Logger, request string) (response string, duratio
 	return response, duration, nil
 }
 
+// repairSourceCodeFile asks the model to repair the code in case any compilation error occurs in a repository source code.
```
bauersimon, line 224: adapt: not `in case` cause we always assume we give the model a proper task, not a task that is incomplete

model/llm/llm_test.go
```diff
@@ -120,3 +122,221 @@ func TestModelGenerateTestsForFile(t *testing.T) {
 		ExpectedTestFilePath: "simple_test.go",
 	})
 }
+
+func TestModelRepairSourceCodeFile(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		SetupMock func(mockedProvider *providertesting.MockQuery)
```
bauersimon, line 130: nit: I know we don't need the `*testing.T` within `SetupMock` right now, but if we do in the future, it is easy to accidentally use the "outside" instance and not the one within `t.Run`. So would be nice to forward it nontheless.

model/llm/llm_test.go
```diff
@@ -120,3 +122,221 @@ func TestModelGenerateTestsForFile(t *testing.T) {
...
+
+		Language       language.Language
+		ModelID        string
```
bauersimon, line 133: Can just set model id to `some-model` within the `validate`. Not something crucial that would change for different test cases.

model/llm/llm_test.go
```diff
@@ -120,3 +122,221 @@ func TestModelGenerateTestsForFile(t *testing.T) {
...
+			` + "```" + `golang
+			package foobar
+			func foobar(i int) int
+				return i + 1
+			}
+			` + "```" + `
+
+			The list of compilation errors is the following:
+
+			- /path/to/foobar.go:3:1: expected 'IDENT', found 'func'
```
bauersimon, line 337: nit: would be cool to remove this new line

task/task.go
```diff
@@ -120,3 +121,115 @@ func (t *TaskWriteTests) Run(repository *repository.Repository) (repositoryAsses
...
+	}()
+
+	filePaths, err := t.Language.Files(log, repository.DataPath)
+	if err != nil {
+		return nil, problems, pkgerrors.WithStack(err)
+	}
```
bauersimon, line 172: For the code repair task, we don't need to go through files, we go through sub-directories and ensure that each subdirectory contains only one implementation file and one test file (and a bunch of others: `go.mod` and `pom.xml`).

task/task.go
```diff
@@ -120,3 +121,115 @@ func (t *TaskWriteTests) Run(repository *repository.Repository) (repositoryAsses
...
+
+	var sourceFilesPath []string
+	for _, filePath := range filePaths {
+		if strings.Contains(filePath, t.Language.DefaultTestFileSuffix()) {
+			continue
+		}
+		if !strings.Contains(filePath, t.Language.DefaultFileExtension()) {
+			continue
+		}
```
bauersimon, line 181: Either use `strings.HasSuffix` or `filepath.Ext` to ensure the extension is in the end of the file path

task/task.go
```diff
@@ -120,3 +121,115 @@ func (t *TaskWriteTests) Run(repository *repository.Repository) (repositoryAsses
...
+		return nil, problems, pkgerrors.Errorf("the repository %q must have only one source file, but found several %#v", repository.DataPath, sourceFilesPath)
+	}
+
+	mistakes, err := t.Language.Mistakes(t.Logger, repository.DataPath)
+	if err != nil {
+		return nil, problems, pkgerrors.WithStack(err)
+	}
+	if len(mistakes) == 0 {
+		return nil, problems, pkgerrors.Errorf("the following repository must contain source files with compilation errors: %q", repository.DataPath)
+	}
```
bauersimon, line 196: We need to go over subdirectories. I am surprised this works, cause within the whole `golang/mistakes` repository, there are 23 go files, so the error should throw every time.

task/task.go
```diff
@@ -120,3 +121,115 @@ func (t *TaskWriteTests) Run(repository *repository.Repository) (repositoryAsses
...
+
+	repositoryAssessment = metrics.NewAssessments()
+
+	if err := repository.Reset(t.Logger); err != nil {
+		t.Logger.Panicf("ERROR: unable to reset temporary repository path: %s", err)
+	}
```
bauersimon, line 202: Please reset the repository in the very beginning (before getting the mistakes).

task/task_test.go
```diff
@@ -88,69 +97,186 @@ func TestTaskWriteTestsRun(t *testing.T) {
...
+				`), nil).After(10 * time.Millisecond) // Simulate a model response delay because our internal safety measures trigger when a query is done in 0 milliseconds.
+			llm := llm.NewModel(mockProvider, modelId)
+
+			validate(t, &testCase{
+				Name: "Go",
+
+				Model:          llm,
+				Language:       &golang.Language{},
+				TestDataPath:   filepath.Join("..", "testdata"),
+				RepositoryPath: filepath.Join("golang", "mistakes", "openingBracketMissing"),
```
bauersimon, line 203: I think the repository is `golang/mistakes`, not `golang/mistakes/openingBracketMissing`. Think back to the `light` repository... with the several examples inside. `openingBracketMissing` is also just one example, not a repository itself. If you want to test this with a real-world testdata example, you need to create an artificial repository with just one example (just make a temp dir for the test and copy the `openingBracketMissing` over there).

bauersimon (changes requested):

testdata/java/mistakes/openingBracketMissing/src/test/java/com/eval/OpeningBracketMissingTest.java
ruiAzevedo19: They'll fail the build since they all have compilation errors

bauersimon: ah yes... well of course they do :face_in_clouds:

evaluate/task/task-write-test.go
```diff
@@ -52,11 +52,6 @@ func (t *TaskWriteTests) Run(repository evaltask.Repository) (repositoryAssessme
 	}
 	defer logClose()
 
-	log.Printf("Evaluating model %q on task %q using language %q and repository %q", t.Model.ID(), t.Identifier(), t.Language.ID(), repository.Name())
-	defer func() {
-		log.Printf("Evaluated model %q on task %q using language %q and repository %q: encountered %d problems: %+v", t.Model.ID(), t.Identifier(), t.Language.ID(), repository.Name(), len(problems), problems)
-	}()
-
```
bauersimon, line 59: Would be cool if we could also include the log file creation, and really move it into a helper cause the evaluation is already soooo cluttered...

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+
+// Run performs source code repairing in a repository with compilation errors.
+func (t *TaskCodeRepair) Run(repository evaltask.Repository) (repositoryAssessment metrics.Assessments, problems []error, err error) {
+	log, logClose, err := log.WithFile(t.Logger, filepath.Join(t.ResultPath, string(t.Identifier()), model.CleanModelNameForFileSystem(t.Model.ID()), t.Language.ID(), repository.Name()+".log"))
+	if err != nil {
+		return nil, nil, err
+	}
+	defer logClose()
+
+	var innerRepositories []string
```
bauersimon, line 61: would call this `taskCases` cause these are the examples we will iterate over

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+	files, err := os.ReadDir(repository.DataPath())
+	if err != nil {
+		return nil, nil, pkgerrors.WithStack(err)
+	}
+	for _, file := range files {
+		if file.IsDir() && !strings.HasPrefix(file.Name(), ".") {
+			innerRepositories = append(innerRepositories, filepath.Join(repository.DataPath(), file.Name()))
```
bauersimon, line 68: What is this prefix check for?

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+		if err := repository.Reset(t.Logger); err != nil {
+			t.Logger.Panicf("ERROR: unable to reset temporary repository path: %s", err)
+		}
+
+		filePaths, err := t.Language.Files(log, innerRepository)
+		if err != nil {
+			return nil, problems, pkgerrors.WithStack(err)
+		}
+		if len(filePaths) != 2 {
+			problems = append(problems, pkgerrors.Errorf("the inner repository must only contain an implementation file and the corresponding test file, but found %#v", filePaths))
```
bauersimon, line 83: I think we should use "task case" for the wording. I know inner "repository" feels tempting cause it is essentially a standalone repository... but to us it is a task case / example / candidate.

bauersimon: Also please report the name of the repository + test case path so we see that during debugging!

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+		var sourceFile string
+		for _, file := range filePaths {
+			if strings.HasSuffix(file, t.Language.DefaultTestFileSuffix()) {
+				hasTestFile = true
+
+				continue
+			}
+			if filepath.Ext(file) == t.Language.DefaultFileExtension() {
+				sourceFile = file
+			}
```
bauersimon, line 97: `else if`

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+		if sourceFile == "" {
+			problems = append(problems, pkgerrors.New("could not find an implementation file"))
+
+			continue
+		}
+		if !hasTestFile {
+			problems = append(problems, pkgerrors.New("could not find a test file"))
+
+			continue
+		}
```
bauersimon, line 108: `else if` and please also put the repository name + test case path here for debugging

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	repositoryAssessment = metrics.NewAssessments()
+	for _, innerRepository := range innerRepositories {
+		if err := repository.Reset(t.Logger); err != nil {
+			t.Logger.Panicf("ERROR: unable to reset temporary repository path: %s", err)
+		}
+
+		filePaths, err := t.Language.Files(log, innerRepository)
```
bauersimon, line 78: Please move all of the validation below into a helper method `validateCodeRepairCase`... might even be worth to somehow just do that once globally at some point. Cause right now we validate the structure for every model and every run even though it does not change. Can you leave/add that as a follow-up please?

bauersimon: Also important cause if the structure is wrong once, then there is somehow no point in continuing the evaluation. Cause it will be wrong in every iteration for every model and run. So checking if all the repository layouts are correct should be something that happens just once in the beginning.

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+			continue
+		}
+
+		mistakes, err := t.Language.Mistakes(t.Logger, innerRepository)
+		if err != nil {
+			problems = append(problems, pkgerrors.WithStack(err))
+
+			continue
+		}
+		if len(mistakes) == 0 {
```
bauersimon, line 116: `else if` since it is still validating the function results

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,157 @@
...
+			problems = append(problems, pkgerrors.Errorf("the following repository must contain source files with compilation errors: %q", repository.DataPath()))
```
bauersimon, line 117: `test case %q in repository %q must contain source files with compilation errors`

this is also an error that we should check once in the very beginning and that should be a hard error cause if something is wrong with the layout, it will stay wrong for the whole evaluation.

evaluate/task/task-code-repair_test.go
```diff
@@ -0,0 +1,269 @@
...
+						return -1
+					}
+
+					return 0
+				}
+			`)
+			modelMock.RegisterGenerateSuccess(t, IdentifierCodeRepair, "openingBracketMissing.go", sourceFileContent, metricstesting.AssessmentsWithProcessingTime).Once()
+
+			validate(t, &tasktesting.TestCaseTask{
+				Name: "Single repository",
```
bauersimon, line 67: `"Single test case"`?

evaluate/task/task-code-repair_test.go
```diff
@@ -0,0 +1,269 @@
...
+						return -1
+					}
+
+					return 0
+				}
+			`)
+			modelMock.RegisterGenerateSuccess(t, IdentifierCodeRepair, "typeUnknown.go", sourceFileContent, metricstesting.AssessmentsWithProcessingTime).Once()
+
+			validate(t, &tasktesting.TestCaseTask{
+				Name: "Multiple repositories",
```
bauersimon, line 130: `"Multiple test cases"`

evaluate/task/task-code-repair_test.go
bauersimon: Same for Java.

evaluate/task/task-code-repair_test.go
bauersimon: Same for Java.

evaluate/task/task-code-repair_test.go
```diff
@@ -0,0 +1,269 @@
...
+	tasktesting "github.com/symflower/eval-dev-quality/evaluate/task/testing"
+	"github.com/symflower/eval-dev-quality/language/golang"
+	"github.com/symflower/eval-dev-quality/language/java"
+	"github.com/symflower/eval-dev-quality/log"
+	modeltesting "github.com/symflower/eval-dev-quality/model/testing"
+	"github.com/zimmski/osutil"
+	"github.com/zimmski/osutil/bytesutil"
+)
+
+func TestTaskCodeRepairRun(t *testing.T) {
```
bauersimon, line 20: These tests are awesome!!! Cannot wait to see the code repair in action :clap:

model/llm/llm_test.go
```diff
@@ -120,3 +123,222 @@ func TestModelGenerateTestsForFile(t *testing.T) {
...
 }
+
+func TestModelRepairSourceCodeFile(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		SetupMock func(t *testing.T, mockedProvider *providertesting.MockQuery)
+
+		Language       language.Language
+		RepositoryPath string
```
bauersimon, line 134: Oh man... this is now also called repository... right... arghhhh... I wanted a clean separation between repository (what contains the test cases) and a test cases itself... But I guess we have to leave it like that in this case cause from the perspective of the LLM this is the repository :see_no_evil:

bauersimon (changes requested):

model/llm/llm.go
bauersimon: unresolved

Second option is great! GitHub deleted my comment when I deleted the duplicate.

evaluate/task/task-code-repair.go
ruiAzevedo19: For hidden directories, like `.git`

evaluate/task/task-write-test.go
ruiAzevedo19: Follow-up

evaluate/task/task-code-repair.go
ruiAzevedo19: Follow-up

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,155 @@
...
+		Logger:     logger,
+	}
+}
+
+// Identifier returns the code repair task identifier.
+func (t *TaskCodeRepair) Identifier() evaltask.Identifier {
+	return IdentifierCodeRepair
+}
+
+// Run performs source code repairing in a repository with compilation errors.
```
ahumenberger, line 53: Please add a docstring saying how the "task cases" are extracted/iterated from a repository.

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, nil, err
+		}
+
+		ctx := evaltask.Context{
+			Language: t.Language,
+
+			RepositoryPath: taskCase,
+			FilePath:       sourceFile,
+
+			Arguments: mistakes,
```
ahumenberger, line 94: `Arguments: &TaskArgumentsCodeRepair{...}`

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, nil, err
+	}
+	defer logClose()
+
+	log.Printf("Evaluating model %q on task %q using language %q and repository %q", t.Model.ID(), t.Identifier(), t.Language.ID(), repository.Name())
+	defer func() {
+		log.Printf("Evaluated model %q on task %q using language %q and repository %q: encountered %d problems: %+v", t.Model.ID(), t.Identifier(), t.Language.ID(), repository.Name(), len(problems), problems)
+	}()
+
+	var taskCases []string
```
ahumenberger, line 66: I find `taskCases` a bit obscure, I don't know what it is. Is it some identifier, is it a path, ....?
One has to read the code below to know what's stored in `taskCases`. Maybe call it `packagePaths`?

model/llm/llm.go
```diff
@@ -103,6 +137,16 @@ func (m *Model) RunTask(ctx task.Context, taskIdentifier task.Identifier) (asses
 	switch taskIdentifier {
 	case evaluatetask.IdentifierWriteTests:
 		return m.generateTestsForFile(ctx)
+	case evaluatetask.IdentifierCodeRepair:
+		mistakes, ok := ctx.Arguments.([]string)
```
ahumenberger, line 141: `taskArgumentsCodeRepair, ok := ctx.Arguments.(*TaskArgumentsCodeRepair)`

ahumenberger (changes requested):

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,155 @@
...
+	ResultPath string
+
+	// Language holds the language for which the task should be evaluated.
+	Language language.Language
+	// Model holds the model which the task should be evaluated.
+	Model model.Model
+
+	// Logger holds the logger for this tasks.
+	Logger *log.Logger
+}
```
ahumenberger, line 28: @ruiAzevedo19 @bauersimon This is not relevant for this PR, but I'd still like to discuss this briefly.

Having this fields in the `TaskXXX` types is pretty unnecessary since we use these fields in a single `Run` method (plus 1-2 helper methods). That is, there is no state that needs to be held.

I was thinking about renaming the current `task.Context` to `task.ModelContext` since the model gets that context, and put the fields of `TaskXXX` into a `task.Context`. and `Task.Run` then gets a `task.Context` as argument.

Thoughts?

bauersimon: sounds good! it is just a collection of arguments after all

ahumenberger: Alright, I'll create a PR based on this one.

ahumenberger (approved):

evaluate/task/task-code-repair.go
ruiAzevedo19: Yap, sounds good to me too

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		log.Printf("Executes tests with %d coverage objects", coverage)
+		repositoryAssessment.Award(metrics.AssessmentKeyFilesExecuted)
+		repositoryAssessment.AwardPoints(metrics.AssessmentKeyCoverage, coverage)
+	}
+
+	return repositoryAssessment, problems, nil
+}
+
+func (t *TaskCodeRepair) validateCodeRepairCase(packagePath string, repository evaltask.Repository, fileLogger *log.Logger) (sourceFile string, mistakes []string, err error) {
```
bauersimon, line 128: Put the logger first cause it is a context argument.

bauersimon, line 128: `sourceFilePath`

bauersimon, line 128: I think it is a good idea to already extract the source file path and mistakes within here. But I think we need a better name for the function, then :thinking:. Cause it is not only validating but also like... "unpacking" the test case... or "extracting"? Please pick what sounds right (maybe you come up with something even better) and also add a docstring.

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+// Identifier returns the code repair task identifier.
+func (t *TaskCodeRepair) Identifier() evaltask.Identifier {
+	return IdentifierCodeRepair
+}
+
+// Run performs source code repairing in a repository with compilation errors.
+// This task requires the repository to consist of multiple packages, with the code repair logic applied to each package.
```
bauersimon, line 54: If we are already going into the details, why not explain it all the way? `consist of multiple packages, with each containing one faulty implementation file and a corresponding test file.`

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,158 @@
...
+	// Model holds the model which the task should be evaluated.
+	Model model.Model
+
+	// Logger holds the logger for this tasks.
+	Logger *log.Logger
+}
+
+// TaskArgumentsCodeRepair holds extra arguments to be used in a query prompt.
+type TaskArgumentsCodeRepair struct {
+	// Mistakes holds the list of compilation errors for a repository.
```
bauersimon, line 32: `package` :angel:

model/llm/llm.go
```diff
@@ -79,6 +79,40 @@ func llmGenerateTestForFilePrompt(data *llmSourceFilePromptContext) (message str
 	return b.String(), nil
 }
 
+// llmCodeRepairSourceFilePromptContext is the template context for a code repair LLM prompt.
+type llmCodeRepairSourceFilePromptContext struct {
+	// llmSourceFilePromptContext holds the context for a source file prompt.
+	llmSourceFilePromptContext
+
+	// Mistakes holds the list of compilation errors of a repository.
```
bauersimon, line 87: `package` :angel:

task/task.go
```diff
@@ -26,6 +26,9 @@ type Context struct {
 	// FilePath holds the path the file under test relative to the repository path.
 	FilePath string
 
+	// Arguments holds extra data that can be used in a query prompt.
+	Arguments any
```
bauersimon, line 30: I think we are almost ready to test-drive this PR. Let me know once everything else is done and we can hop on a call to try it out with some real models.

bauersimon (changes requested):

evaluate/task/task-code-repair.go
bauersimon: Please add a docstring :pray:

evaluate/task/task-code-repair.go
```diff
@@ -0,0 +1,158 @@
...
+		return "", nil, pkgerrors.WithStack(err)
+	} else if len(mistakes) == 0 {
+		return "", nil, pkgerrors.Errorf("package %q in repository %q must contain source files with compilation errors", packagePath, repository.Name())
+	}
+
+	filePaths, err := t.Language.Files(fileLogger, packagePath)
+	if err != nil {
+		return "", nil, pkgerrors.WithStack(err)
+	}
+	if len(filePaths) != 2 {
```
bauersimon, line 140: `else if ` :sob: :bow:

bauersimon (changes requested):

language/java/language.go
```diff
@@ -129,7 +139,12 @@ func (l *Language) Mistakes(logger *log.Logger, repositoryPath string) (mistakes
 	})
 	if err != nil {
 		if output != "" {
-			return extractMistakes(output), nil
+			mistakes := extractMistakes(output)
+			for i, mistake := range mistakes {
+				mistakes[i] = strings.TrimPrefix(mistake, repositoryPath)
+			}
+
+			return mistakes, nil
```
bauersimon, line 147: Add comment.

I guess we will have the same problem with windows again where the reported maven compile errors are with unix slashes. Not sure if worth to look into.

bauersimon (changes requested):

bauersimon:
todo solution for `variableUnknown` is ambiguous

PR #171, opened 2024-06-10
fix, Fail tests immediately in case tool is outdated or unusable

bauersimon:


cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -91,6 +92,11 @@ func validateMetrics(t *testing.T, regex *regexp.Regexp, data string, expectedAs
 }
 
 func TestEvaluateExecute(t *testing.T) {
+	if osutil.IsLinux() {
```
zimmski, line 95: @ahumenberger it would be better to have a testing solution that is in one place instead of multiple. If you have ideas. At the moment, i don't

bauersimon: yeah... I tried to come with one but also... kinda out of ideas...

tools/testing/tools_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	validate := func(t *testing.T, tc *testCase) {
+		t.Run(tc.Name, func(t *testing.T) {
+			tt := &testing.T{}
+			mock := NewMockToolNamed(t, "some-tool")
+			tc.SetupMock(t, mock)
+
+			// The "testing" logic internally calls "runtime.Goexit" in case of a failure which exits a Goroutine (but still runs "defer"s) so start a new one to not kill ourselves.
+			var waitGroup sync.WaitGroup
```
zimmski, line 29: Why do we need a waitgroup? This code does not start a goroutine and IMO it shouldn't in general start one. That would be surprising.

tools/testing/tools_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/mock"
+
+	"github.com/symflower/eval-dev-quality/tools"
+)
+
+func TestRequiresTool(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		SetupMock    func(t *testing.T, m *MockTool)
```
zimmski, line 18: Convention: Group into Name, inputs, outputs

tools/testing/tools_test.go
```diff
@@ -0,0 +1,67 @@
...
+		ExpectFailed bool
+	}
+
+	validate := func(t *testing.T, tc *testCase) {
+		t.Run(tc.Name, func(t *testing.T) {
+			tt := &testing.T{}
+			mock := NewMockToolNamed(t, "some-tool")
+			tc.SetupMock(t, mock)
+
+			// The "testing" logic internally calls "runtime.Goexit" in case of a failure which exits a Goroutine (but still runs "defer"s) so start a new one to not kill ourselves.
```
ahumenberger, line 28: `// The "testing" package calls "runtime.Goexit" in case of a failure which exits the innermost Goroutine. We therefore call the function under test in a new Goroutine such that the outer one gets killed.

tools/testing/tools_test.go
bauersimon: Explanation is right above. If we have a test case where `RequiresTool` fails and calls `FailNow`, that will internally call `runtime.Goexit`. That would kill "ourselves" (I had that already, that's why I introduced the goroutine). So we need a sub-goroutine that is killed instead.

bauersimon: Also... this code **does** start a goroutine... right below :thinking:

cmd/eval-dev-quality/cmd/evaluate_test.go
bauersimon: Only way I could think of would be to introduce a `testing.T`  wrapper that we need to use everywhere instead of the normal one. And that one could then probably to the tools check automatically every time.

ahumenberger (changes requested):

cmd/eval-dev-quality/cmd/evaluate_test.go
zimmski: > Only way I could think of would be to introduce a `testing.T` wrapper that we need to use everywhere instead of the normal one. And that one could then probably to the tools check automatically every time.

Too complicated.

Another solution is to have a testing package and do it there in an "init" function. Would lead to having all these checks only once.

tools/testing/tools_test.go
zimmski: > Also... this code **does** start a goroutine... right below 🤔

The testing code you mean?

zimmski: > Explanation is right above. If we have a test case where `RequiresTool` fails and calls `tt.FailNow`, that will internally call `runtime.Goexit`. That would kill "ourselves" (I had that already, that's why I introduced the goroutine). So we need a sub-goroutine that is killed instead.

When you have code that does "exit" calls, test from outside. Not in the same process.

cmd/eval-dev-quality/cmd/evaluate_test.go
bauersimon: where would this testing package be? or do you mean moving all the `_test.go` files into a separate package? I hope that we never access any private things cause otherwise that would not be doable.

tools/testing/tools_test.go
bauersimon: I really think that is overly complicated... `testing.T.FailNow` calls `runtime.Goexit` but that is an internal technicality that we unfortunately run into here. Is it really worth it making this a system tests that includes writing a file and calling `go test`? I guess alternatively I could mock `testing.T`... a bit less complex.

cmd/eval-dev-quality/cmd/evaluate_test.go
zimmski: Let's rework this another time.

tools/testing/tools_test.go
bauersimon: obsolete

tools/testing/tools_test.go
bauersimon: @zimmski plz take another look at this

ahumenberger: Looks good to me.

ahumenberger (approved):

PR #172, opened 2024-06-10
fix, Use a backoff for retrying LLM queries because it seems that some LLMs need longer to recover

zimmski:


model/llm/llm.go
```diff
@@ -126,9 +126,6 @@ func (m *Model) GenerateTestsForFile(logger *log.Logger, language language.Langu
 		retry.OnRetry(func(n uint, err error) {
 			logger.Printf("Attempt %d/%d: %s", n+1, m.queryAttempts, err)
 		}),
-		retry.RetryIf(func(err error) bool {
-			return true
-		}),
```
bauersimon, line 131: `refactor,` commit cause behavior stays?

model/llm/llm.go
```diff
@@ -122,6 +122,7 @@ func (m *Model) GenerateTestsForFile(logger *log.Logger, language language.Langu
 		},
 		retry.Attempts(m.queryAttempts),
 		retry.Delay(5*time.Second),
+		retry.DelayType(retry.BackOffDelay),
```
bauersimon, line 125: Not sure if I would call this a `fix` cause... we cannot know if it really fixes... but fine with me anyways cause there is no test case.

model/llm/llm.go
zimmski: yes

model/llm/llm.go
zimmski: Looking at the logs, it should.

zimmski: We cannot test this problem automatically because we cannot introduce a failure in the API providers. This is what the data shows, this would have fixed what the data shows. Internally, this is a fix.

model/llm/llm.go
bauersimon: think you did not push this?

bauersimon: its not important... will merge

PR #177, opened 2024-06-11
fix, Allow arbitrary content immediately after code tag

bauersimon:


model/llm/prompt/parse_test.go
```diff
@@ -107,16 +107,29 @@ func TestParseResponse(t *testing.T) {
...
+				Response: "Some text...\n\n```\n" + code + "\n```\n\nSome more text...",
+
+				ExpectedAssessment: metrics.Assessments{
+					metrics.AssessmentKeyResponseNoExcess: 0,
+					metrics.AssessmentKeyResponseWithCode: 1,
+				},
+				ExpectedCode: code,
+			})
+			validate(t, &testCase{
+				Name: "No Newline",
```
ahumenberger, line 123: The examples of the issue have `.` right before the code fence. Is this a typo, or actually the response we get? If so, there is no test case for that, right? Like no new line before the code fence.

ahumenberger (changes requested):

model/llm/prompt/parse_test.go
bauersimon: I used that to mark the space character because it is very easy to miss. It should say so in the issue. The actual example had a space before the first fence, and we already have that example.

ahumenberger (approved):

PR #179, opened 2024-06-11
fix, Rename result directory if it already exists

bauersimon:


cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -131,7 +131,7 @@ func TestEvaluateExecute(t *testing.T) {
 
 			arguments := append([]string{
 				"evaluate",
-				"--result-path", temporaryPath,
+				"--result-path", filepath.Join(temporaryPath, "result-directory"),
```
ruiAzevedo19, line 134: What's the reason for adding `result-directory` everywhere?

ruiAzevedo19:
Should we change the milestone to v6?

cmd/eval-dev-quality/cmd/evaluate_test.go
bauersimon: With this change the results are moved somewhere else in case the results directory already exists. In the tests we create a temporary directory and give it to the evaluation. And since the temporary directory **exists**, all results are placed somewhere else. So instead I tell the evaluation to put the results in `<temporaryDirectory>/result-directory`. And since `result-directory` does not exist within the temporary directory yet, everything finally lands in there. Only in the final new test case I actually create `result-directory` and observe that the results are instead placed into `result-directory-0` :see_no_evil:

bauersimon: So basically if we hand the evaluation just the temporary directory that we create for the tests, it will not write anything in there because the temporary directory already exists ^^... so I need to add another subdirectory inside that does not exist yet.

bauersimon: I thought about solving this somehow without having to change all test cases... but that just means I have to introduce some hidden magic somewhere that works around that and that will just confuse us in the future.

ruiAzevedo19: Sounds good to me

PR #180, opened 2024-06-12
ffix, Create temporary repositories once

### LINKED ISSUE #157: Logic for "Create temporary repositories for each language so the repository is copied only once per language."  copies more than needed
The logic for reusing copied repositories should only copy repositories that are actually used. When filtered on e.g. only java/plain, the java/light repository is copied as well. That is a flawed logic because that makes debugging harder. Every repository that we add will make that worse. Debugging will get harder.

bauersimon:
Fixes #157


ruiAzevedo19:
@bauersimon Looks good

PR #182, opened 2024-06-12
fix, Default to all repositories if none are selected in CLI

bauersimon:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -205,25 +214,45 @@ func (command *Evaluate) Initialize(args []string) (cleanup func()) {
...
+			for languageID := range languagesSelected {
+				if commandRepositoriesLanguages[languageID] {
+					commandRepositories[filepath.Join(languageID, evaluate.RepositoryPlainName)] = true
+				} else {
+					command.Languages = slices.DeleteFunc(command.Languages, func(l string) bool {
+						return l == languageID
+					})
+					delete(languagesSelected, languageID)
+					command.logger.Printf("Excluded language %q because it is not part of the selected repositories", languageID)
+				}
```
ahumenberger, line 255: I find that a bit complicated. And we are having two different computation paths for basically the same situation: Assume the user does not specify a repository, that is, we do the `if` branch with the result that all repositories are evaluated. But if the user specifies all repositories, we do the `else` branch, again with the result that all repositories are evaluated. So we use two different paths for the same thing.

Ideally it should be something like this:

```go
if len(command.Repositories) == 0 {
  command.Repositories = AllRepositories
}

if len(command.Languages) == 0 {
  command.Languages == AllLanguages
}

if len(command.Tasks) == 0 {
  command.Tasks = AllTasks
}

ValidateOptions()
```

ahumenberger (changes requested):

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -255,44 +287,51 @@ func (command *Evaluate) Execute(args []string) (err error) {
...
 
-	ls := make([]language.Language, len(command.Languages))
-	for i, languageID := range command.Languages {
-		ls[i] = languagesSelected[languageID]
-	}
-	ms := make([]model.Model, len(command.Models))
-	for i, modelID := range command.Models {
-		ms[i] = modelsSelected[modelID]
-	}
 	assessments, totalScore := evaluate.Evaluate(&evaluate.Context{
```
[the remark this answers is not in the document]
bauersimon: huuuuuuuh.... yeeeeeeees :thinking:

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: Sure. For languages and tasks it is kinda straight forward (though the problem with languages is that we need to update them according to which repositories are selected).

The only problem is that `AllRepositories` means going through `testdata` and figuring out which ones there are (which is exactly what I am doing in the `if` case`). So yes we use two different paths for the same thing, but the first one is us reading the directory to figure out which repositories there are and the second one is just handling the users input, so two different scenarios.

ahumenberger: > though the problem with languages is that we need to update them according to which repositories are selected).

Should be the same with tasks at some point I think.

The difference between your appraoch an mine is that mine always performs option validation, and it's a separate step and easier to maintain IMO.
Your approach does not validate options if there are no repositories specified, because we do not have to yet. But once we have a better integration of tasks there needs to be a validation as well.

What I don't like about that code (not just the one you are changing) is that initialization and validation is mixed. I'd like to have validation in a separate function.

But okay, I guess it's fine for now. But please move the `if` branch into a funciton, and move `commandRepositories` and `commandRepositoriesLanguages` into the branches. They are not need afterwards AFAIS.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -79,29 +81,51 @@ func (command *Evaluate) SetLogger(logger *log.Logger) {
...
+				cleanup()
+			}
+			panic(r)
+		}
+	}()
+	evaluationContext = &evaluate.Context{}
+
+	// Setup evaluation result directory.
+	{
+		command.timestamp = time.Now()
```
ahumenberger, line 99: I'd move this to the beginning of `Evaluate.Execute` as before, I think.

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -191,6 +191,8 @@ func TestEvaluateExecute(t *testing.T) {
 			Arguments: []string{
 				"--language", "golang",
 				"--model", "symflower/symbolic-execution",
+				"--repository", filepath.Join("golang", "plain"),
+				"--repository", filepath.Join("java", "plain"),
```
ahumenberger, line 195: Why is `java/plain` listed if the language is `golang`?

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -918,4 +925,25 @@ func TestEvaluateInitialize(t *testing.T) {
...
+			directories, err := os.ReadDir(filepath.Join("..", "..", "..", "testdata", "golang"))
+			require.NoError(t, err)
+			relativeRepositoryPath := make([]string, len(directories))
+			for i, directory := range directories {
+				relativeRepositoryPath[i] = filepath.Join("golang", directory.Name())
+			}
+
+			for _, golangRepository := range relativeRepositoryPath {
+				assert.Contains(t, command.Repositories, golangRepository)
+			}
```
ahumenberger, line 946: What about Java repositories?

ahumenberger (changes requested):

cmd/eval-dev-quality/cmd/evaluate_test.go
bauersimon: Sure... but do we want to change this test case every time we add a new language? I think I even mentioned in comment that this is just a sanity check.

cmd/eval-dev-quality/cmd/evaluate_test.go
bauersimon: errrrrrr... yes I think we can get rid of that one...

cmd/eval-dev-quality/cmd/evaluate_test.go
ahumenberger: Feels wrong to do a test for "selecting no repository defaults to all" and then we just check a few repositories. Added a loop to iterate over `language.Languages`.

ahumenberger (approved):

PR #184, opened 2024-06-13
Explicit windows test case for Java test path logic

### LINKED ISSUE #159: https://github.com/symflower/eval-dev-quality/pull/155/files missing a test
All change need tests in TDD style. fail first, pass with changes. Same rules as internally.

bauersimon:
Closes #159

### DIFF
--- a/language/java/language_test.go
+++ b/language/java/language_test.go
@@ -113,13 +113,23 @@ func TestLanguageTestFilePath(t *testing.T) {
 		})
 	}
 
-	validate(t, &testCase{
-		Name: "Source file",
+	if osutil.IsWindows() {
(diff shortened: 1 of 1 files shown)


language/java/language_test.go
```diff
@@ -113,13 +113,23 @@ func TestLanguageTestFilePath(t *testing.T) {
 		})
 	}
 
-	validate(t, &testCase{
-		Name: "Source file",
+	if !osutil.IsWindows() {
```
Munsio, line 116: Just a nitpick but would remove the negation and swap the condition content. When skimming over code it is easier having a `true` case.

PR #188, opened 2024-06-13
fix, Define a timeout for the "symflower test" command, so ensure the execution does not take too much time

ruiAzevedo19:


language/language.go
```diff
@@ -1,11 +1,17 @@
...
 import (
+	"time"
+
 	pkgerrors "github.com/pkg/errors"
 
 	"github.com/symflower/eval-dev-quality/log"
 )
 
+// DefaultExecutionTimeout defines the timeout for an execution.
+// WORKAROUND For now we define the timeout as a global variable but must be moved to the "symflower test" command.
```
bauersimon, line 12: `but it should eventually be moved...`

util/exec_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/symflower/eval-dev-quality/log"
+)
+
+func TestCommandWithResult(t *testing.T) {
```
bauersimon, line 12: Would rename to `TestCommandWithResultTimeout` because it tests one one specific behavior.

util/exec_test.go
```diff
@@ -0,0 +1,33 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(1*time.Second))
+	defer cancel()
+
+	start := time.Now()
+	CommandWithResult(ctx, logger, &Command{
+		Command: []string{
+			"sleep",
+			"60",
+		},
+	})
```
bauersimon, line 29: Please assert that there is an error returned so we know that hitting the timeout produces an error.

util/exec_test.go
```diff
@@ -0,0 +1,33 @@
...
+	duration := time.Since(start)
+
+	assert.Less(t, duration.Seconds(), time.Duration(5*time.Second).Seconds())
```
bauersimon, line 32: I think you can just change `time.Duration(.......` to `5` cause... 5 seconds have 5 seconds :blush:

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -59,6 +59,8 @@ type Evaluate struct {
 	// TestdataPath determines the testdata path where all repositories reside grouped by languages.
 	TestdataPath string `long:"testdata" description:"Path to the testdata directory where all repositories reside grouped by languages." default:"testdata/"`
 
+	// ExecutionTimeout holds the timeout for an execution.
+	ExecutionTimeout int `long:"execution-timeout" description:"Execution timeout in minutes" default:"0"`
```
bauersimon, line 63: Please always make full sentences. "Execution timeout for compilation and tests in minutes.".

Also, I just realized that the default value we have is also printed in the help text. So we should use 5 here - sorry :see_no_evil:. Also please use `uint` to make sure it cannot be negative.

bauersimon (changes requested):

bauersimon (approved):

PR #189, opened 2024-06-13
Script for sequentially evaluating common models with "light" repository

bauersimon:


scripts/evaluation/sequential-light.go
```diff
@@ -0,0 +1,92 @@
+package main
+
```
ahumenberger, line 2: looks fine. but I'd rephrase the commit message to fit our conventions and avoid "Add". 

Suggestion: `Script for sequentially evaluating common models with "light" repository`

PR #192, opened 2024-06-14
Early merger for code repair task

ruiAzevedo19:
Part of #168


model/llm/llm.go
```diff
@@ -90,6 +90,8 @@ func (m *Model) IsTaskSupported(taskIdentifier task.Identifier) (isSupported boo
 	switch taskIdentifier {
 	case task.IdentifierWriteTests:
 		return true
+	case task.IdentifierCodeRepair:
+		return true
```
bauersimon, line 94: well... kinda not supported yet...

PR #202, opened 2024-06-19
Share logging setup between tasks

bauersimon:


evaluate/task/task.go
```diff
@@ -47,3 +50,35 @@ func TaskForIdentifier(taskIdentifier evaltask.Identifier) (task evaltask.Task,
...
 	}
 }
+
+// taskLogging holds common logging functionality.
+type taskLogging struct {
+	log      *log.Logger
+	logClose func()
+	ctx      evaltask.Context
+	task     evaltask.Task
+}
```
ahumenberger, line 60:
```go
type taskLogger struct {
	*log.Logger

	logClose func()
	ctx      evaltask.Context
	task     evaltask.Task
}
```

Then you still have "initialize" and "finalize" but use the struct as regular logger.
Then maybe rename "initializeLogging" to `NewTaskLogger`?!

ahumenberger (changes requested):

evaluate/task/task.go
bauersimon: Will not export though to prevent misuse.

ahumenberger (approved):

PR #203, opened 2024-06-19
refactor, Move fields of "TaskWriteTests" and "TaskCodeRepair" into context object since the tasks itself are stateless

ahumenberger:


bauersimon (changes requested):

task/identifier/identifier.go
```diff
@@ -0,0 +1,11 @@
+package identifier
```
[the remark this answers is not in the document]
ahumenberger: obsolete

model/model.go
```diff
@@ -13,7 +16,24 @@ type Model interface {
...
+}
+
+// Context holds the data needed by a model for running a task.
+type Context struct {
+	// Language holds the language for which the task should be evaluated.
+	Language language.Language
+
+	// RepositoryPath holds the absolute path to the repository.
+	RepositoryPath string
+	// FilePath holds the path the file under test relative to the repository path.
```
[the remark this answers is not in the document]
ahumenberger: `// FilePath holds the path to the file the model should act on.`

?

evaluate/evaluate_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	modeltesting "github.com/symflower/eval-dev-quality/model/testing"
 	"github.com/symflower/eval-dev-quality/provider"
 	providertesting "github.com/symflower/eval-dev-quality/provider/testing"
-	"github.com/symflower/eval-dev-quality/task"
+	"github.com/symflower/eval-dev-quality/task/identifier"
```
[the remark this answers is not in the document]
ahumenberger: Got rid of `Model.RunTask` and introduced separate "capability" interfaces which the models need to implement:

```go
// CapabilityWriteTests defines the capability of a model to generate tests.
type CapabilityWriteTests interface {
	Model

	// WriteTests generates test files for the given implementation file in a repository.
	WriteTests(ctx Context) (assessments metrics.Assessments, err error)
}

// CapabilityRepairCode defines the capability of a model to repair code.
type CapabilityRepairCode interface {
	Model

	// RepairCode queries the model to repair a source code with compilation error.
	RepairCode(ctx Context) (assessments metrics.Assessments, err error)
}
```

Also interesting for @ruiAzevedo19 I suppose.

model/capability.go
```diff
@@ -0,0 +1,19 @@
...
+	WriteTests(ctx Context) (assessments metrics.Assessments, err error)
+}
+
+// CapabilityRepairCode defines the capability of a model to repair code.
+type CapabilityRepairCode interface {
+	Model
+
+	// RepairCode queries the model to repair a source code with compilation error.
+	RepairCode(ctx Context) (assessments metrics.Assessments, err error)
+}
```
bauersimon, line 19: I think you can also remove the `Model` embedding. Cause this is something that anyways just happens "ontop" of a model.

evaluate/task/testing/task.go
```diff
@@ -11,15 +11,14 @@ import (
 	"github.com/symflower/eval-dev-quality/evaluate/metrics"
 	metricstesting "github.com/symflower/eval-dev-quality/evaluate/metrics/testing"
 	"github.com/symflower/eval-dev-quality/language"
-	"github.com/symflower/eval-dev-quality/model"
 	evaltask "github.com/symflower/eval-dev-quality/task"
 	"github.com/zimmski/osutil"
 )
 
-type TestCaseTask struct {
+type TestCaseTask[ModelCapability any] struct {
```
bauersimon: https://go.dev/doc/tutorial/generics#declare_type_constraint ?

evaluate/task/test-integration/task_test.go
```diff
@@ -22,22 +21,14 @@ func TestTaskWriteTestsRun(t *testing.T) {
...
-
-			tc.Validate(t, taskWriteTests, repository, resultPath, logger)
+			tc.Validate(t,
+				func(logger *log.Logger, testDataPath string, repositoryPathRelative string) (repository evaltask.Repository, cleanup func(), err error) {
+					return evaluatetask.TemporaryRepository(logger, testDataPath, repositoryPathRelative)
+				},
+				func() (task evaltask.Task, err error) {
+					return evaluatetask.TaskForIdentifier(evaluatetask.IdentifierWriteTests)
+				},
+			)
```
bauersimon, line 31: I guess we still cannot move this entirely into the `testing`  package cause we would have a cyclic import between testing and implementation :disappointed_relieved:

bauersimon: but we also cannot move them into one cause we want the `testing` package to use it in the `integration` tests... grrrrrr. At this point... why not just have it in one single package? It all tests the same thing? Forcing it to be split is just super annoying at this point.

bauersimon (changes requested):

model/capability.go
ahumenberger: True.

evaluate/task/testing/task.go
ahumenberger: Does not work with interfaces having methods

evaluate/task/test-integration/task_test.go
ahumenberger: I could get rid of the second callback. the first one is still there. We need to keep refactoring to get rid of that.

bauersimon: Would just chuck everything into one package cause it is all testing the task... But it's already better than before.

bauersimon (approved):

evaluate/task/test-integration/task_test.go
ahumenberger: > Would just chuck everything into one package cause it is all testing the task... But it's already better than before.

That would just hide the fact that our non-testing package structure is still not perfect. But we are getting there.

PR #211, opened 2024-06-24
Docker runtime

Munsio:


Munsio:
@bauersimon can you do a review + early merger?

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -361,6 +362,8 @@ func (command *Evaluate) Execute(args []string) (err error) {
 	switch command.Runtime {
 	case "local":
 		return command.localRuntime(evaluationContext)
+	case "docker":
+		return command.dockerRuntime(evaluationContext)
```
bauersimon, line 372: This is great! Two things that come to mind:

- [ ] we automatically install i.e. `ollama` and `symflower` so should we automatically install `docker` as well?
  - I think ultimately: yes - but that can be a follow up
- [ ] we should add tests for this that run symflower in a docker container on the golang/plain (I guess we have to manually install docker in the CI for now)

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -158,6 +160,10 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 		}
 		evaluationContext.Runs = command.Runs
 
+		if command.RuntimeImage == "" {
+			command.RuntimeImage = "ghcr.io/symflower/eval-dev-quality:v" + evaluate.Version
+		}
```
bauersimon, line 165: awesome! Maybe some tests as well? Just using a plain alpine image and ensure it fails?

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -440,3 +444,69 @@ func (command *Evaluate) localRuntime(evaluationContext *evaluate.Context) (err
...
+
+		cmd := append(dockerCommand, evaluationCommand...)
+		cmd = append(cmd, args...)
+
+		commandOutput, err := util.CommandWithResult(context.Background(), command.logger, &util.Command{
+			Command: cmd,
+		})
+		if err != nil {
+			return pkgerrors.WithMessage(pkgerrors.WithStack(err), commandOutput)
+		}
```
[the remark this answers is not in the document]
Munsio: `CommandWithResult` blocks with `docker run` until the run is finished and then continues with the next one.
Parallel runs is a new iteration.

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: Installing docker is not as trivial as install symflower or ollama, alone the Network stack problem that is happening with some configurations. 
Forcing to install docker/ollama on a user is something I do not like. If I only want to run openrouter models I don't need anything.
We should provide the right errors + maybe ask if the user wants that we install some stuff (except symflower as it is a hard dependency) 

Also "running" docker commands inside github actions isn't easy by itself, sure there is https://github.com/addnab/docker-run-action but that would run a job inside a container but we want to test our code which uses the "docker" shell command to run a container which is totally different.

util/exec_test.go
```diff
@@ -32,3 +32,41 @@ func TestCommandWithResultTimeout(t *testing.T) {
...
+		},
+		Ignored: []string{
+			"--runtime",
+		},
+
+		ExpectedFiltered: []string{
+			"--runs",
+			"5",
+		},
+	})
```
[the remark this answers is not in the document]
Munsio: Ah damn why doesn't `os.Args` already solve that problem for me. Back to the drawing board.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -68,6 +68,9 @@ type Evaluate struct {
 	// NoDisqualification indicates that models are not to be disqualified if they fail to solve basic language tasks.
 	NoDisqualification bool `long:"no-disqualification" description:"By default, models that cannot solve basic language tasks are disqualified for more complex tasks. Overwriting this behavior runs all tasks regardless."`
 
+	// Runtime indicates if the evaluation is run locally or inside a container.
+	Runtime string `long:"runtime" description:"The runtime which will be used for the evaluation." default:"local"`
```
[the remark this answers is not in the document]
Munsio: Ah yeah "choice:`bla`" nice - didn't know that.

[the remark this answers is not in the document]
ruiAzevedo19: Took a look at `jessevdk/go-flags` documentation and there is a `choice` option:

`long:"..." description:"..." choice:"local" choice:"docker" default:"local"`

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -440,3 +444,69 @@ func (command *Evaluate) localRuntime(evaluationContext *evaluate.Context) (err
...
+		}
+
+		// Create for each model a dedicated subfolder inside the results path.
+		resultPath, err := filepath.Abs(command.ResultPath)
+		if err != nil {
+			return err
+		}
+		if err := osutil.MkdirAll(resultPath); err != nil {
+			return err
+		}
```
ruiAzevedo19: Isn't this creating the same folder over again? Shouldn't we join the model ID to the result path?

util/exec_test.go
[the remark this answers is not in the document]
Munsio: No because `--foo=bar` is one entry in the slice and not two.

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: Yeah that is outdated sry `MkdirAll` isn't necessary because we already created it. But what we need is `chmod` it to 777 - as we are running `evalaute` again inside the container with the result-path set there it is creating the folders accoridng to the subcommand.

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: Follow up so we know this is untested for now.

util/exec.go
```diff
@@ -54,3 +54,39 @@ func CommandWithResult(ctx context.Context, logger *log.Logger, command *Command
...
+	filterMap := map[string]bool{}
+	for _, v := range ignored {
+		filterMap[v] = true
+	}
+
+	// Resolve args with equals sign.
+	var resolvedArgs []string
+	for _, v := range args {
+		if strings.HasPrefix(v, "--") && strings.Contains(v, "=") {
+			resolvedArgs = append(resolvedArgs, strings.Split(v, "=")...)
```
bauersimon, line 69: Maybe `SplitN(v, "=", 1)`

Munsio: :eagle: :eye:

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: Don't get what you really want? I run it already multiple times as I had to verify it is working. The only thing you could do is checking out the branch and run it yourself! That would be a successful test.

Here is the command I used for testing:
```
make install && eval-dev-quality evaluate --runtime docker --result-path ./docker-test --runs 5 --model symflower/symbolic-execution --model symflower/symbolic-execution --model symflower/symbolic-execution --repository golang/plain
```

Munsio: Also here is the follow up issue: https://github.com/symflower/eval-dev-quality/issues/224

bauersimon (approved):

PR #218, opened 2024-06-25
Multiple Runs again

Munsio:


ahumenberger:
@Munsio Do we still need this PR?

bauersimon:
not part of an official evaluation ... we can still reopen and merge if we feel like it

Munsio:
Would ask @zimmski if he wants the results in any reporting folder inside the Repo, also thought that those runs are just some kind of "lets try that out"

PR #221, opened 2024-06-26
Parallel execution of containerized evaluations

Munsio:


Munsio:
@bauersimon I would like to have your opinion on this one.

bauersimon:
plz rebase and then tag me again

Munsio:
Alright I got it to work. I also updated the follow up issue already to also add tests for `--parallel` when we run docker tests on the CI.

Please check.

Also if you have a idea for testing the `Parallel` stuff please tell me.

Also here is the command I used to run:
```

File: util/exec.go [hunk 1]
```diff
@@ -90,3 +91,43 @@ func FilterArgs(args []string, ignored []string) (filtered []string) {
...
 }
+
+// Parallel holds a buffered channel for limiting parallel executions.
+type Parallel struct {
+	ch chan struct{}
+	wg sync.WaitGroup
+}
+
+// NewParallel returns a Parallel execution limiting helper.
+func NewParallel(limit uint, timeout time.Duration) *Parallel {
```
bauersimon, line 102: Would maybe call that `maxWorkers`.

util/exec_test.go
```diff
@@ -91,3 +91,39 @@ func TestFilterArgs(t *testing.T) {
...
+		Name: "Run 10 times with limit of 2",
+
+		Timeout: 0,
+		Limit:   2,
+		Times:   10,
+
+		Function: func() {
+			// do something?
+		},
+	})
```
bauersimon, line 128: Would have one test case where every function adds `+1` to some mutex and then you check if really all 10x of them ran.
And a second case maybe where you have `time.Sleep(1)` in every function with `Limit:1` and `Times:3` and make sure the whole thing runs for at least like `2.5` seconds.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -74,6 +74,8 @@ type Evaluate struct {
 	Runtime string `long:"runtime" description:"The runtime which will be used for the evaluation." default:"local" choice:"local" choice:"docker"`
 	// RuntimeImage determines the container image used for any container runtime.
 	RuntimeImage string `long:"runtime-image" description:"The container image to use for the evaluation." default:""`
+	// Parallel holds the number of parallel executed runs.
+	Parallel uint `long:"parallel" description:"Amount of parallel executed runs." default:"1"`
```
bauersimon, line 78: State that this only affects docker.

util/exec.go
bauersimon, line 102: Timeout parameter unused.

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: Though I would maybe make the logic different (maybe in a follow up). Cause now we have runs and parallel executions... Would be cool if `parallel` would be a boolean and it automatically distributes multiple runs across parallel executions?

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: Yeah but runs and parallel are different.
For better result spread I want to have 5 runs of each model, while my server is able to run 15-20 parallel evaluations.

Making it exclusive for containerized runs and clarify it in the comment is the way to go I guess.

util/exec_test.go
Munsio: The first one sounds right and enough, I do not want to use timed counts with `sleep` as they artificially increase execution time where it is not needed, as in fact both do the same thing only the measurement is different.

util/exec.go
Munsio: Ah yeah forgot to remove, thanks.

Munsio:
@bauersimon please review again.

util/exec_test.go
bauersimon: mhmmm... if you don't want sleeps than I am not sure how to easily measure that only x run at a time.

util/exec_test.go
```diff
@@ -91,3 +92,35 @@ func TestFilterArgs(t *testing.T) {
...
+
+		p := NewParallel(tc.Limit)
+		for i := uint(0); i < tc.Times; i++ {
+			p.Execute(tc.Function)
+		}
+		p.Wait()
+
+		assert.Equal(t, count.Load(), uint32(expectedCount))
+	})
+
```
bauersimon, line 125: Unused newline.

util/exec_test.go
```diff
@@ -91,3 +92,35 @@ func TestFilterArgs(t *testing.T) {
...
 	})
 }
+
+func TestParallelExecute(t *testing.T) {
+	type testCase struct {
+		Limit uint
+		Times uint
+
+		Function func()
+	}
```
bauersimon, line 102: Kinda weird having table tests that are not used in any table fashion. Would drop them.

util/exec_test.go
```diff
@@ -91,3 +92,35 @@ func TestFilterArgs(t *testing.T) {
...
+
+	t.Run("Atomic count", func(t *testing.T) {
```
bauersimon, line 104: So this ensures that the functions are executed correctly. Not sure how to test that the limit works without using `time.Sleep`. I guess you could make something that adds and subtracts from atomic and also checks what the maximum value was. And with a limit of 1 then the maximum value should never exceed 2. But as long as it is functionally correct, we are good I think.

bauersimon (dismissed):

util/exec_test.go
Munsio: Hmm yeah you are right, testing the limit is only possible with time. Fixed that up. Please review again @bauersimon

util/exec_test.go
bauersimon: Still weird cause we construct them within the tests and then just use their properties. So would either introduce a more elaborate `validate` or really just remove the `testCase`. To me it makes everything harder to read.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -144,6 +146,10 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 			}
 		}
 
+		if command.Parallel != 1 && command.Runtime == "local" {
+			command.logger.Panic("the 'parallel' parameter can't be used with local execution")
+		}
+
```
bauersimon, line 152: `command.Parallel` could be `0` ... sry for not noticing earlier

Munsio:
As discussed with @bauersimon everything is ready to merge with the fix of the pipeline.

Merge it right away

PR #227, opened 2024-06-28
(minimal )asdf support for running the evaluations

kohlerm:
I am using http://asdf-vm.com/ for my dev environment setup. 
It picks up the runtimes from the .tool-versions file. 
Compilation works fine but running the eval did not work because the tests are done in a temporary directory which do not have a .tool-versions file.
this attempts to fix this by just copying the .tool-versions file the temp folder. 
Works fine for me on MacOS, tests with some local ollama model.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ go.work
 
 # Ignore local configurations.
 .envrc
+
+# Ignore the evaluation directories
+/evaluation-*
--- a/evaluate/task/repository.go
+++ b/evaluate/task/repository.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


bauersimon:
@kohlerm with the addition of Docker, it is now not recommended to run evaluations "bare-metal". And the Docker image will contain all the tools anyways. Do you think it is still worth pursuing this? I think we would just add a new flag that allows you to specify patterns of files to always copy over into the temporary directories.

kohlerm:
Agreed, avoiding it to run it bare metal is a good idea. Your proposal is fine with me.

bauersimon:
Will close this PR then. Thanks so much for your contribution regardless. And let us know if anything else comes up in the future.

Issue #213, opened 2024-06-25
Apply symflower fix to a "write-test" result of a model


bauersimon:
Basically we want to execute the "write-test" task, but then optionally call `symflower repair` on the generated tests.
Plus, the scoring should treat both the original "write-test" and the (hopefully) repaired tests as different results.

- [x] introduce a new `TaskIdentifier` `write-test-symflower-repair` that represents the write test task but with symflower applied
- [x] let `task.Run` return `map[TaskIdentifier]Assessments` such that it can return both the unfixed, and fixed results
- [x] extend the "write-test" task such that
  - [x] if the first result fails at execution apply `symflower repair` and executes again
  - [x] write the results of `symflower repair` into the original model's log, i.e. `with symflower repair: .......`
  - [x] watch out for the scoring... the score must be combined intelligently
    - runtime probably has to be added up
    - for coverage and execution only the final "attempt" matters
    - ...
- [x] `symflower fix` needs to be merged

bauersimon:
@Munsio plz review

Munsio:
- When we want to "repair" something don't we need the data from the previous model? How is it shared?
- How does the `Next` Model in the chain know that it needs to repair something and not generate the thing new?

bauersimon:
The task evaluation logic can carry over the "broken code" and just ask the next model to do a "repair". The models themselves don't need to worry about sharing context or what they need to do. The evaluation logic will say "you failed, but you, please fix, this is what we have so far".

PR #229, opened 2024-06-28
Apply "symflower fix" to a "write-test" result of a model when it errors, so model responses can possibly be fixed

ruiAzevedo19:
Part of #213

### DIFF
--- a/evaluate/evaluate.go
+++ b/evaluate/evaluate.go
@@ -126,7 +126,7 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore, totalScore uin
 								}
 
 								assessment, ps, err := task.Run(temporaryRepository)
-								assessments.Add(model, language, repositoryPath, taskIdentifier, assessment)
+								assessments.AddAssessmentPerTask(model, language, repositoryPath, assessment)
 								if err != nil {
 									ps = append(ps, err)
 								}
@@ -226,7 +226,7 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore, totalScore uin
 								}
 
 								assessment, ps, err := task.Run(temporaryRepository)
-								assessments.Add(model, language, repositoryPath, taskIdentifier, assessment)
+								assessments.AddAssessmentPerTask(model, language, repositoryPath, assessment)
 								problemsPerModel[modelID] = append(problemsPerModel[modelID], ps...)
 								if err != nil {
 									ctx.Log.Printf("ERROR: Model %q encountered a hard error for language %q, repository %q: %+v", modelID, languageID, repositoryPath, err)
--- a/evaluate/evaluate.go
+++ b/evaluate/evaluate.go
@@ -249,7 +249,8 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore, totalScore uin
 		}
 	}
 	if isOnlyPlainRepositories {
-		totalScore = uint64(len(ctx.Languages)) * uint64(ctx.Runs)
+		// For every write-test task in the plain repository, each model is also executed with the `symflower fix` which results in double the total results.
+		totalScore = 2 * uint64(len(ctx.Languages)) * uint64(ctx.Runs)
 	}
 
 	return assessments, totalScore
--- a/evaluate/metrics/testing/assessments.go
+++ b/evaluate/metrics/testing/assessments.go
@@ -1,10 +1,12 @@
 package metricstesting
 
 import (
-	"maps"
 	"testing"
 
+	"golang.org/x/exp/maps"
+
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/symflower/eval-dev-quality/evaluate/metrics"
 	"github.com/symflower/eval-dev-quality/language"
@@ -17,15 +19,39 @@ func AssertAssessmentsEqual(t *testing.T, expected metrics.Assessments, actual m
 	expected = maps.Clone(expected)
 	actual = maps.Clone(actual)
 
-	expected[metrics.AssessmentKeyProcessingTime] = 0
-	actual[metrics.AssessmentKeyProcessingTime] = 0
-
-	expected[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
-	actual[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
-	expected[metrics.AssessmentKeyResponseCharacterCount] = 0
-	actual[metrics.AssessmentKeyResponseCharacterCount] = 0
+	clearNonDeterministicAssessmentValues(expected)
+	clearNonDeterministicAssessmentValues(actual)
 
 	assert.Truef(t, expected.Equal(actual), "expected:%s\nactual:%s", expected, actual)
+}
+
+// AssertTaskAssessmentsEqual checks if the given assessments per task are equal ignoring default and nondeterministic values.
+func AssertTaskAssessmentsEqual(t *testing.T, expected map[task.Identifier]metrics.Assessments, actual map[task.Identifier]metrics.Assessments) {
+	expected = maps.Clone(expected)
+	actual = maps.Clone(actual)
+
+	// The expected and actual maps must have the same task identifiers.
+	require.ElementsMatch(t, maps.Keys(expected), maps.Keys(actual))
+
+	// Ignore non-deterministic values.
+	for _, assessment := range expected {
+		clearNonDeterministicAssessmentValues(assessment)
+	}
+	for _, assessment := range actual {
+		clearNonDeterministicAssessmentValues(assessment)
+	}
+
+	for task, expectedAssessment := range expected {
+		actualAssessment := actual[task]
+		assert.Truef(t, expectedAssessment.Equal(actualAssessment), "task:%s\nexpected:%s\nactual:%s", task, expected, actual)
+	}
+}
+
+// clearNonDeterministicAssessmentValues ignores non-deterministic values such as processing time and response character count.
+func clearNonDeterministicAssessmentValues(assessment metrics.Assessments) {
+	assessment[metrics.AssessmentKeyProcessingTime] = 0
+	assessment[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
+	assessment[metrics.AssessmentKeyResponseCharacterCount] = 0
 }
 
 // AssessmentsWithProcessingTime is an empty assessment collection with positive processing time.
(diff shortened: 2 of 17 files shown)


evaluate/report/collection.go
```diff
@@ -84,6 +84,12 @@ func (a *AssessmentStore) Add(model model.Model, l language.Language, repository
 	assessments.Add(assessment)
 }
 
+func (a *AssessmentStore) AddAssessmentPerTask(model model.Model, l language.Language, repositoryPath string, taskAssessment map[task.Identifier]metrics.Assessments) {
```
bauersimon, line 87: Docstring missing

evaluate/metrics/testing/assessments.go
```diff
@@ -28,6 +31,46 @@ func AssertAssessmentsEqual(t *testing.T, expected metrics.Assessments, actual m
...
+
+			return false
+		})
+	}
+	expectedTaskIdentifiers := maps.Keys(expected)
+	sortedTaskIdentifier(expectedTaskIdentifiers)
+	actualTaskIdentifiers := maps.Keys(actual)
+	sortedTaskIdentifier(actualTaskIdentifiers)
+
+	require.Equal(t, expectedTaskIdentifiers, actualTaskIdentifiers)
```
bauersimon, line 54: Can use `require.ElementsMatch` to avoid all the sorting business.

evaluate/metrics/testing/assessments.go
```diff
@@ -28,6 +31,46 @@ func AssertAssessmentsEqual(t *testing.T, expected metrics.Assessments, actual m
...
+	for _, assessment := range expected {
+		assessment[metrics.AssessmentKeyProcessingTime] = 0
+		assessment[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
+		assessment[metrics.AssessmentKeyResponseCharacterCount] = 0
+	}
+	for _, assessment := range actual {
+		assessment[metrics.AssessmentKeyProcessingTime] = 0
+		assessment[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
+		assessment[metrics.AssessmentKeyResponseCharacterCount] = 0
+	}
```
bauersimon, line 66: Please share code with the other assertion function.

evaluate/metrics/testing/assessments.go
```diff
@@ -28,6 +31,46 @@ func AssertAssessmentsEqual(t *testing.T, expected metrics.Assessments, actual m
...
+
+	for task, expectedAssessment := range expected {
+		actualAssessment, _ := actual[task] // At this point they have the same task identifiers.
```
bauersimon, line 69: Can just use `actualAssessment := actual[task]`.

bauersimon (changes requested):

evaluate/metrics/testing/assessments.go
```diff
@@ -17,17 +19,48 @@ func AssertAssessmentsEqual(t *testing.T, expected metrics.Assessments, actual m
...
+	// actual[metrics.AssessmentKeyProcessingTime] = 0
 
-	expected[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
-	actual[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
-	expected[metrics.AssessmentKeyResponseCharacterCount] = 0
-	actual[metrics.AssessmentKeyResponseCharacterCount] = 0
+	// expected[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
+	// actual[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
+	// expected[metrics.AssessmentKeyResponseCharacterCount] = 0
+	// actual[metrics.AssessmentKeyResponseCharacterCount] = 0
```
bauersimon, line 31: errrrrr no comments in code that we want to merge

evaluate/metrics/testing/assessments.go
```diff
@@ -17,17 +19,48 @@ func AssertAssessmentsEqual(t *testing.T, expected metrics.Assessments, actual m
...
+		assert.Truef(t, expectedAssessment.Equal(actualAssessment), "task:%s\nexpected:%s\nactual:%s", task, expected, actual)
+	}
+}
+
+// ignoreNonDeterministicAssessmentValues ignores non-deterministic values such as processing time and response character count.
+func ignoreNonDeterministicAssessmentValues(assessment metrics.Assessments) {
+	assessment[metrics.AssessmentKeyProcessingTime] = 0
+	assessment[metrics.AssessmentKeyGenerateTestsForFileCharacterCount] = 0
+	assessment[metrics.AssessmentKeyResponseCharacterCount] = 0
+}
```
bauersimon, line 62: `clearNonDeterministicAssessmentValues`

evaluate/task/symflower-fix.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// The processing time is summed with the one from the model.
+	assessments[metrics.AssessmentKeyProcessingTime] = modelAssessment[metrics.AssessmentKeyProcessingTime] + uint64(duration.Milliseconds())
+
+	coverage, ps, err := language.Execute(logger, repositoryPath)
+	problems = append(problems, ps...)
+	if err != nil {
+		return nil, ps, pkgerrors.WithStack(err)
+	}
+	logger.Printf("with symflower repair: Executes tests with %d coverage objects", coverage)
```
bauersimon, line 43: I think this should still be part of the task code. Cause symflower fix should not care what happens with its result.

evaluate/task/symflower-fix.go
```diff
@@ -0,0 +1,50 @@
...
+	pkgerrors "github.com/pkg/errors"
+	"github.com/symflower/eval-dev-quality/evaluate/metrics"
+	"github.com/symflower/eval-dev-quality/language"
+	"github.com/symflower/eval-dev-quality/log"
+	"github.com/symflower/eval-dev-quality/tools"
+	"github.com/symflower/eval-dev-quality/util"
+)
+
+// symflowerFix runs the "symflower fix" command and returns the corresponding assessments.
+func symflowerFix(logger *log.Logger, modelAssessment metrics.Assessments, repositoryPath string, language language.Language) (assessments metrics.Assessments, problems []error, err error) {
```
bauersimon, line 16: Please split this function to have the symflower fix command call separated form the assessment combining logic. And consider moving the assessment combination logic into the assessment package where it can easily be unit tested (i.e. if the model had this assessment and the fixed result had that assessment then result should be...).

evaluate/task/symflower-fix.go
```diff
@@ -0,0 +1,50 @@
...
+			"--workspace", repositoryPath,
+		},
+
+		Directory: repositoryPath,
+	})
+	if err != nil {
+		return nil, nil, pkgerrors.WithStack(err)
+	}
+	duration := time.Since(start)
+	logger.Printf("with symflower repair: %s", output)
```
bauersimon, line 33: don't think we need this log output cause the command logic is logging everything on its own

evaluate/task/task-write-test.go
```diff
@@ -93,7 +94,18 @@ func (t *TaskWriteTests) Run(repository evaltask.Repository) (repositoryAssessme
 		if err != nil {
 			problems = append(problems, pkgerrors.WithMessage(err, filePath))
 
-			continue
+			// Run "symflower fix"  if the model response fails to execute.
+			if t.Language.ID() == "golang" { // Currently we only support Go for "symflower fix".
+				assessments, ps, err := symflowerFix(log, modelAssessment, dataPath, t.Language)
```
bauersimon, line 99: plz add a log i.e. (model response alone failed execution, attempting to fix with `symflower fix`)

bauersimon (changes requested):

evaluate/metrics/testing/assessments.go
ruiAzevedo19: Missed that, I thought I have deleted those

evaluate/task/task-write-test.go
```diff
@@ -85,19 +86,37 @@ func (t *TaskWriteTests) Run(repository evaltask.Repository) (repositoryAssessme
...
-		repositoryAssessment.AwardPoints(metrics.AssessmentKeyCoverage, coverage)
+		modelAssessment.Award(metrics.AssessmentKeyFilesExecuted)
+		modelAssessment.AwardPoints(metrics.AssessmentKeyCoverage, coverage)
+	}
+
+	repositoryAssessment = map[evaltask.Identifier]metrics.Assessments{
+		IdentifierWriteTests: modelAssessment,
+	}
+	if len(symflowerAssessment) > 0 {
+		repositoryAssessment[IdentifierWriteTestsSymflowerFix] = symflowerAssessment
```
bauersimon, line 119: I guess what we want is that the symflower assessment is the non-fix assessment in case we did not have to fix.

I.e. such that if we look at the final result for `write-test-symflower-fix` we don't want to spend time to figure out which cases worked fine

evaluate/task/task-write-test.go
```diff
@@ -43,7 +43,7 @@ func (t *TaskWriteTests) Identifier() evaltask.Identifier {
 }
 
 // TaskWriteTests generates test files for the given implementation file in a repository.
-func (t *TaskWriteTests) Run(repository evaltask.Repository) (repositoryAssessment metrics.Assessments, problems []error, err error) {
+func (t *TaskWriteTests) Run(repository evaltask.Repository) (repositoryAssessment map[evaltask.Identifier]metrics.Assessments, problems []error, err error) {
```
ahumenberger, line 46: Have you thought about passing the assessment store to `TaskWriteTests.Run` as parameter, and letting the the assessments be stored directly in there instead of returning a map of assessments?

I introduced the assessment store to get rid of the `map[x]map[y]map[z]....`. It feels a bit like we are starting that again: starting with a single level map, but then we need more and it becomes again a `map[x]map[y]map[z]...` thing.

This is nothing you need to change now if this already works, but something to keep in mind please.

bauersimon: Think we tried that but Rui got a cyclic import.

ruiAzevedo19:
Follow-up:  Don't run `symflower fix` if there is a timeout on the LLM response

evaluate/evaluate.go
```diff
@@ -249,7 +249,8 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore, totalScore uin
 		}
 	}
 	if isOnlyPlainRepositories {
-		totalScore = uint64(len(ctx.Languages)) * uint64(ctx.Runs)
+		// For each task, the `symflower fix` is also called so we multiply the total score by 2.
```
bauersimon, line 252: For every write-test task in the plain repository, each model is also executed with `symflower fix` which results in double the total results.

bauersimon:
@ruiAzevedo19 LGTM plz take a look as well at my changes

ruiAzevedo19:
@bauersimon Also LGTM

bauersimon (approved):

Issue #207, opened 2024-06-21
Add the current commit revision to the binary, Docker image and reports


zimmski:
We currently have the problem that we sometimes do not know if a report is using the correct binary / repository revision.

Tasks:
- [x] Add revision and repository version to
  - [x]  Binary
  - [x] Docker image
  - [x] Reports

PR #234, opened 2024-07-02
Docker runtime follow up

Munsio:


Munsio:
Currently the CI job itself is flaky with some permission problems. I will investigate that one further but in the meantime the PR can be reviewed.

Munsio:
We need to override the docker image with the tag from the current branch.
Also we need to run the tests after the image was created.

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -95,6 +98,14 @@ func (command *Evaluate) SetLogger(logger *log.Logger) {
 	command.logger = logger
 }
 
+var _ SetArguments = (*Evaluate)(nil)
+
+// SetArguments sets the commands arguments.
+func (command *Evaluate) SetArguments(args []string) {
+	availableFlags := util.Flags(command)
+	command.args = util.FilterArgsKeep(args, availableFlags)
+}
```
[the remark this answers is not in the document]
Munsio: cmd/eval-dev-quality/cmd/command.go

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -768,6 +769,175 @@ func TestEvaluateExecute(t *testing.T) {
...
+
+		validate(t, &testCase{
+			Name: "Docker",
+
+			Arguments: []string{
+				"--runtime", "docker",
+				"--model", "symflower/symbolic-execution",
+				"--model", "symflower/symbolic-execution",
+				"--model", "symflower/symbolic-execution",
+				"--testdata", "testdata/", // We need to override the testdata path back to the default one because the actual evaluation is running inside the container.
```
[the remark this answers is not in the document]
Munsio: Because it gets set inside the `validation` function `cmd/eval-dev-quality/cmd/evaluate_test.go:140` it is a default which is used everywhere except for the containerized stuff because we do not have any temporary testdata path there but the testdata from the image itself.

cmd/eval-dev-quality/cmd/evaluate.go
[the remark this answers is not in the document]
bauersimon: So all of this is basically to avoid `os.Args[2:]`?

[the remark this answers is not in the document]
Munsio: Yes because `os.Args` contains all the flags used for testing too etc. and because of that it was not possible to get a clear slice of the parameters which should be passed down to the container image.

cmd/eval-dev-quality/cmd/evaluate_test.go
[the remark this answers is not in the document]
bauersimon: Maybe then change the docstring to `Our own tests set the "testdata" argument to the temporary directory that they create. This temporary directory does not exist in docker, so set the "testdata" manually here to overrule the testing behavior and use the original one.`

Munsio:
@bauersimon docker test runs through. Please review all again did some major reworking.

README.md
```diff
@@ -135,9 +135,16 @@ Total coverage 100.000000%
...
 The execution by default also creates an report file `REPORT.md` that contains additional evaluation results and links to individual result files.
 
-# Docker
+# Containerized use
 
-## Setup
+## Notes
+
+The following parameters do have a special behavior when using a containerized runtime.
+- `--testdata`: The check if the path exists is ignored on the host system but still enforced inside the container.
```
bauersimon, line 143: `because the paths of the host and inside the container might differ.`

Dockerfile
```diff
@@ -1,40 +1,62 @@
...
-# Setup the evaluation
+# Install the binary
+COPY --from=builder --chown=ubuntu:ubuntu /app/eval-dev-quality /app/.eval-dev-quality/bin/
+ENV PATH="${PATH}:/app/.eval-dev-quality/bin"
+RUN go install -v github.com/vektra/mockery/v2@v2.40.3 && go install -v gotest.tools/gotestsum@v1.11.0
+RUN eval-dev-quality install-tools --install-tools-path /app/.eval-dev-quality/bin
 
-RUN make install-all
-ENV PATH="${PATH}:/home/ubuntu/.eval-dev-quality/bin"
+# CHMOD everything because of non-root user ids
```
bauersimon, line 61: Bit inconvenient that we install here manually instead of using the makefile. Cause then we have to maintain two things. IS there a way to refactor the makefile to have everything there still?

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -448,18 +459,31 @@ func (command *Evaluate) evaluateLocal(evaluationContext *evaluate.Context) (err
...
+
+	// Get current user for volume ID mapping.
+	user, err := user.Current()
+	if err != nil {
+		return pkgerrors.WithStack(err)
+	}
+
+	resultPath, err := filepath.Abs(command.ResultPath)
+	if err != nil {
+		return err
```
bauersimon, line 481: `pkgerrors`

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -448,18 +459,31 @@ func (command *Evaluate) evaluateLocal(evaluationContext *evaluate.Context) (err
...
+	}
+	// Set permission 777 so the non-root docker image is able to store its results inside the result path.
+	if err := os.Chmod(resultPath, 0777); err != nil {
+		return err
```
bauersimon, line 485: `pkgerrors`

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -768,6 +771,244 @@ func TestEvaluateExecute(t *testing.T) {
...
 
+	t.Run("Runtime", func(t *testing.T) {
+		// Skip containerized runtime tests if special cases.
+		if osutil.IsWindows() {
+			t.Skip("Docker runtime not supported on Windows")
+		} else if _, err := exec.LookPath("docker"); err != nil {
+			t.Skip("Docker runtime not found")
+		}
+
+		// Prerequisite
```
bauersimon, line 782: Extend docstring to that we pull the image of the current branch.

Could we not just build the image here to be 100% sure it is the right one?

Munsio: The CI is built that docker build is a preqreuisite step before the tests run. I am not sure if we should put a docker image build into the testing!?

bauersimon: ok... if we 100% have the correct image here than it's fine... just afraid of having some networking hickups and then this image is some previous one

Dockerfile
Munsio: Split it up into two commands where one of it takes an argument, building the container works as well as running `make install-all`

Munsio:
I had to skip the docker runtime in local tests for now as we do have a problem with the "commit hashes" - I hope we can fix that with: #207 

Docker runtime tests are skipped locally and only run inside the Github Actions CI

Issue #232, opened 2024-07-02
Follow-up: Apply "symflower fix" to a "write-test" result of a model when it errors, so model responses can possibly be fixed


ruiAzevedo19:
### TODO

- [x] Don't run symflower fix if there is a timeout on the LLM response https://github.com/symflower/eval-dev-quality/pull/229#issuecomment-2202912134

PR #236, opened 2024-07-02
Do not run "symflower fix" if the original response failed with a timeout, so the model and the fix assessments are consistent

ruiAzevedo19:
Part of #232

### DIFF
--- a/evaluate/task/task-write-test.go
+++ b/evaluate/task/task-write-test.go
@@ -1,6 +1,8 @@
 package task
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 
@@ -81,6 +83,14 @@ func (t *TaskWriteTests) Run(ctx evaltask.Context) (repositoryAssessment map[eva
 		problems = append(problems, ps...)
 		if err != nil {
 			problems = append(problems, pkgerrors.WithMessage(err, filePath))
+
+			// If there is an execution timeout do not run "symflower fix" because the code itself is correct.
+			if errors.Is(err, context.DeadlineExceeded) {
+				modelAssessment.Add(modelAssessmentForFile)
+				withSymflowerAssessment.Add(withSymflowerAssessmentForFile)
+
+				continue
+			}
 
 			// Run "symflower fix"  if the model response fails to execute.
 			if ctx.Language.ID() == "golang" { // Currently we only support Go for "symflower fix".
--- a/evaluate/task/task-write-test_test.go
+++ b/evaluate/task/task-write-test_test.go
@@ -1,18 +1,23 @@
 package task
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


ruiAzevedo19:
@bauersimon I wanted to add a unit test for this case but I think it's not that easy:
- the execution timeout value is hard-coded, so in order to test the changes I would need to change `language.DefaultExecutionTimeout = <timeout-for-tests>` and then reset it to the original value. This approach sounds wrong
- I would need a scenario where the 1st call to the execution function (for the model)  times out, but the 2nd call (for `symflower fix`) does not. Not possible with the current implementation
- Also thought of a system test where we can set the execution time via CLI, but the value is in minutes.

evaluate/task/task-write-test.go
```diff
@@ -97,6 +99,14 @@ func (t *TaskWriteTests) Run(repository evaltask.Repository) (repositoryAssessme
 		if err != nil {
 			problems = append(problems, pkgerrors.WithMessage(err, filePath))
 
+			// If there is an execution timeout do not run "symflower fix".
```
bauersimon, line 102: `cause the code itself is correct`.

bauersimon:
I think one clean possibility would be to have the execution timeout as a method of the language interface and then use a mock of a language that can return a lower value. Can you try that out and if it takes longer then 10 minutes we just go with what we have? It's just an optimization for like a 1% chance. So not super duper important to test if it's such a hassle.

ruiAzevedo19:
> I think one clean possibility would be to have the execution timeout as a method of the language interface and then use a mock of a language that can return a lower value. Can you try that out and if it takes longer then 10 minutes we just go with what we have? It's just an optimization for like a 1% chance. So not super duper important to test if it's such a hassle.

Did it a bit differently: I mocked the `Execute` function to return the timeout error and did the assertions accordingly. Reason being the timeout method would be called inside the `Execute` method, which needs also to be mocked.

evaluate/task/task-write-test_test.go
```diff
@@ -178,12 +182,42 @@ func TestTaskWriteTestsRun(t *testing.T) {
...
+				}
+				expectedProblems := []string{
+					"context deadline exceeded",
+				}
+
+				languageMock := languagetesting.NewMockLanguageNamed(t, "golang")
+				languageMock.On("Files", mock.Anything, mock.Anything).Return([]string{filepath.Join("golang", "plain")}, nil).Once()
+				languageMock.On("Execute", mock.Anything, mock.Anything).Return(uint64(0), nil, context.DeadlineExceeded).Once()
+
+				validateGo(t, "Model generated test with unused import", languageMock, bytesutil.StringTrimIndentations(`
```
bauersimon, line 208: update test name to reflect timeout

evaluate/task/task-write-test_test.go
```diff
@@ -178,12 +182,42 @@ func TestTaskWriteTestsRun(t *testing.T) {
...
+					package plain
+
+					import (
+						"testing"
+						"strings"
+					)
+
+					func TestPlain(t *testing.T) {
+					   	plain()
+					}
```
bauersimon, line 218: Please make this an empty string so it's clear that it is not used.

bauersimon (changes requested):

bauersimon (approved):

Issue #237, opened 2024-07-03
Dump the assessments in the CSV files once they happen and not in the end of all executions


ruiAzevedo19:
## TODO 

**1st iteration**
- [x] Dump the assessments into the `evaluation.csv`  every time a task is executed

**2nd iteration**
- [x] Create the other CSVs from the `evaluation.csv`
  - read `evaluation.csv` > sum the results > write to file
  - [x] `models-summed.csv`
  - [x] `<language>-summed.csv` 

### Follow-ups
- [x] Store the tasks and repositories in the evaluation records  
  - https://github.com/symflower/eval-dev-quality/pull/241#discussion_r1666425640
- [ ] Add the model human-readable name and its costs only when the raw records of the `evaluation.csv` are processed 
  - https://github.com/symflower/eval-dev-quality/pull/241#discussion_r1666436342
- [x] Change `EvaluationRecordsPerModel` to be `map[string][]*metrics.Assessments`
  - https://github.com/symflower/eval-dev-quality/pull/241#discussion_r1668526526

bauersimon:
- would basically just have only one `evaluation.csv` where everything is just dumped on the fly
- the other two csvs can be created from the first one, it should just be different summing
  - thats also already the first step towards "combining multiple evaluations"
  - basically one single "raw" results file, and then we can do all the pretty stuff in a separate step
- basically in this iteration you just read in the raw csv and do the summing up and write it out separately

bauersimon:
remaining task will be implemented with https://github.com/symflower/eval-dev-quality/issues/296

PR #240, opened 2024-07-04
Dump the assessments to the evaluation CSV right after running a task, to avoid losing information if the evaluation errors

ruiAzevedo19:
Part of #237

PR #241, opened 2024-07-04
Generate the "models-summed.csv" and "language-summed.csv" files based on the "evaluation.csv" file

ruiAzevedo19:
Part of #237


ruiAzevedo19:
Based on https://github.com/symflower/eval-dev-quality/pull/240

evaluate/evaluate_test.go
```diff
@@ -180,6 +180,8 @@ func TestEvaluate(t *testing.T) {
 			Before: func(t *testing.T, logger *log.Logger, resultPath string) {
 				// Set up mocks, when test is running.
 				mockedModel.MockCapabilityWriteTests.On("WriteTests", mock.Anything).Return(nil, ErrEmptyResponseFromModel)
+				mockedModel.MockModel.On("Name").Return("Empty Response Model")
+				mockedModel.MockModel.On("Cost").Return(0.0001)
```
bauersimon, line 184: please put that in the helper that creates the mock since we will have it all the time

evaluate/report/csv.go
```diff
@@ -26,50 +27,109 @@ type CSVFormatter interface {
...
-	}
+	// LanguageID holds the language id.
+	LanguageID string
 
-	for _, row := range formatter.Rows() {
-		if err := csv.Write(row); err != nil {
-			return "", pkgerrors.WithStack(err)
+	// Assessments holds the assessments of an entry.
+	Assessments metrics.Assessments
+}
```
bauersimon, line 44: This should probably also hold the task and repository, right?

evaluate/report/csv.go
```diff
@@ -26,50 +27,109 @@ type CSVFormatter interface {
...
-	var out strings.Builder
-	csv := csv.NewWriter(&out)
+// EvaluationRecord holds a line of the evaluation CSV.
+type EvaluationRecord struct {
+	// ModelID holds the model id.
+	ModelID string
+	// ModelName holds the model name.
+	ModelName string
+	// ModelCost holds the model cost.
+	ModelCost float64
```
bauersimon, line 37: Since those things never change (during an evaluation)... makes me wonder if we just move them at some point such that they are added when the raw records are processed. And not pulled along the whole time.

evaluate/report/csv.go
```diff
@@ -121,3 +181,164 @@ func WriteEvaluationRecord(resultPath string, model model.Model, language langua
 
 	return nil
 }
+
+// loadEvaluationRecords reads and returns the evaluation records from the evaluation CSV file.
+func loadEvaluationRecords(resultPath string) (evaluationRecords EvaluationRecords, err error) {
+	evaluationFilePath := filepath.Join(resultPath, "evaluation.csv")
```
bauersimon, line 187: would not hand the result path but only the file where the raw records are. less hardcoded values

evaluate/report/csv.go
```diff
@@ -121,3 +181,164 @@ func WriteEvaluationRecord(resultPath string, model model.Model, language langua
...
+		return nil, pkgerrors.WithStack(err)
+	}
+	defer evaluationFile.Close()
+
+	reader := csv.NewReader(evaluationFile)
+
+	// Check if the evaluation CSV header is correct.
+	if header, err := reader.Read(); err != nil {
+		return nil, pkgerrors.Wrap(err, "file is empty")
+	} else if !reflect.DeepEqual(header, EvaluationHeader()) {
```
bauersimon, line 199: This should be two strings, no?

evaluate/report/csv.go
```diff
@@ -121,3 +181,164 @@ func WriteEvaluationRecord(resultPath string, model model.Model, language langua
...
+	if err != nil {
+		return nil, pkgerrors.WithStack(err)
+	}
+	assessments[metrics.AssessmentKeyResponseNoExcess] = responseNoExcess
+
+	responseWithCode, err := strconv.ParseUint(raw[14], 10, 64)
+	if err != nil {
+		return nil, pkgerrors.WithStack(err)
+	}
+	assessments[metrics.AssessmentKeyResponseWithCode] = responseWithCode
```
bauersimon, line 281: Please just do a `for` loop since all the values are `uint`. The order of the keys can be obtained from the assessment package. It's the same order as it was written. That way it is also futureproof in case we change the order of keys or add new ones.

evaluate/report/csv.go
```diff
@@ -26,50 +27,109 @@ type CSVFormatter interface {
...
+		} else {
+			r := perModel[record.ModelID]
+			r.Assessments = metrics.Merge(r.Assessments, record.Assessments)
 		}
 	}
 
-	csv.Flush()
-
-	return out.String(), nil
+	return perModel
```
bauersimon, line 79: Since you moved some grouping functionality here, I would think that the assessment package now has some functions that are unused, right? Can you please check? I would think that the only reason we have all the assessment magic is the categorization. Which we should either remove completely or move to here at some point.

evaluate/report/csv_test.go
```diff
@@ -198,3 +219,602 @@ func TestWriteEvaluationRecord(t *testing.T) {
 		`,
 	})
 }
+
+func TestLoadEvaluationRecords(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		Before func(reaultPath string)
```
bauersimon, line 227: typo

evaluate/report/csv.go
```diff
@@ -121,3 +181,164 @@ func WriteEvaluationRecord(resultPath string, model model.Model, language langua
...
+	if err != nil {
+		return nil, pkgerrors.WithStack(err)
+	}
+	defer evaluationFile.Close()
+
+	reader := csv.NewReader(evaluationFile)
+
+	// Check if the evaluation CSV header is correct.
+	if header, err := reader.Read(); err != nil {
+		return nil, pkgerrors.Wrap(err, "file is empty")
```
bauersimon, line 198: is it always empty? would maybe say `malformed`?

bauersimon (changes requested):

evaluate/report/csv.go
ruiAzevedo19: My 1st implementation held both the task and repository fields. The reasons I removed them are:
- I'm using the evaluation record to group the values and adding the assessments, so what would be e.g the `Repository` value when  we group two evaluation records from different repositories? 
- We don't need them for the current implementation

If we want to still keep them I guess we need another struct similar to `EvaluationRecord` but without the tasks and repositories, for grouping the values.

evaluate/report/csv.go
ruiAzevedo19: I guess it makes sense just adding the costs when we process the `evalution.csv`. AFAIK we don't need the costs during the evaluation, so that's something we can do afterwards.  

Can be done in a follow-up

evaluate/report/csv.go
ruiAzevedo19: The `Read` function returns the header as a string slice, same type as `EvaluationHeader`

evaluate/report/csv.go
ruiAzevedo19: Tried writing random data to the file but no error like that. Still, I changed the error message to be more generic

evaluate/report/csv.go
ruiAzevedo19: Also, something we can do in a follow-up if needed

evaluate/report/csv.go
ruiAzevedo19: Removed the unused functions/structs. There are still some things used in the markdown report and in the evaluation itself.

evaluate/report/csv.go
bauersimon: maybe let's just do `strings.Join` and then compare the strings. Calling `reflect` feels a bit overkill

evaluate/report/csv.go
bauersimon: Cannot wait to get rid of all of that as well.

evaluate/report/csv.go
bauersimon: Plz note down in the issue and resolve this.

evaluate/report/csv.go
bauersimon: Plz note down in the issue and resolve this.

evaluate/report/csv.go
ruiAzevedo19: Good call

ruiAzevedo19:
@bauersimon Please, take a final look

evaluate/evaluate.go
```diff
@@ -75,11 +75,12 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore, totalScore uin
...
 	problemsPerModel := map[string][]error{}
 	// Write the evaluation CSV header so it's only written once.
-	file, err := os.OpenFile(filepath.Join(ctx.ResultPath, "evaluation.csv"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	evaluationCSVFile, err := os.OpenFile(filepath.Join(ctx.ResultPath, "evaluation.csv"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
 	if err != nil {
 		ctx.Log.Panicf("ERROR: unable to create evaluation CSV file: %+v", err)
 	}
-	evaluationFile, err := report.NewEvaluationFile(file)
+	defer evaluationCSVFile.Close()
+	evaluationFile, err := report.NewEvaluationFile(evaluationCSVFile)
```
bauersimon, line 83: Rename could probably go into previous commit.

evaluate/report/csv.go
```diff
@@ -124,3 +82,228 @@ func (e *EvaluationFile) WriteEvaluationRecord(model model.Model, language langu
...
+
+	rawMetrics := raw[7:]
+	for i, assessementKey := range metrics.AllAssessmentKeysStrings {
+		metric, err := strconv.ParseUint(rawMetrics[i], 10, 64)
+		if err != nil {
+			return nil, pkgerrors.WithStack(err)
+		}
+
+		assessments[metrics.AssessmentKey(assessementKey)] = metric
+	}
```
bauersimon, line 246: nitpick, would be nice to check if the length matches :see_no_evil:

bauersimon (changes requested):

evaluate/report/csv.go
ruiAzevedo19: I don't think we need it, the `reader.ReadAll()` in  `loadEvaluationRecords` already checks it the records have the length.

evaluate/report/csv.go
```diff
@@ -88,3 +68,59 @@ func (a AssessmentPerModel) Rows() (rows [][]string) {
...
+	csv := csv.NewWriter(&out)
+
+	if err := csv.Write(evaluationHeader()); err != nil {
+		return nil, pkgerrors.WithStack(err)
+	}
+	csv.Flush()
+
+	if _, err = evaluationFile.Writer.Write([]byte(out.String())); err != nil {
+		return nil, pkgerrors.WithStack(err)
+	}
```
bauersimon: Sorry for noticing just now. You should be able to avoid the `strings.Builder` and just do `csv.NewWriter(writer)`.

evaluate/report/csv.go
```diff
@@ -88,3 +68,59 @@ func (a AssessmentPerModel) Rows() (rows [][]string) {
...
+
+	return evaluationFile, nil
+}
+
+// WriteEvaluationRecord writes the assessments of a task into the evaluation CSV.
+func (e *EvaluationFile) WriteEvaluationRecord(model model.Model, language language.Language, repositoryName string, assessmentsPerTask map[task.Identifier]metrics.Assessments) (err error) {
+	var out strings.Builder
+	csv := csv.NewWriter(&out)
```
bauersimon, line 60: Sorry for noticing just now. You should be able to avoid the `strings.Builder` and just do `csv.NewWriter(writer)`.

evaluate/report/csv.go
```diff
@@ -124,3 +82,228 @@ func (e *EvaluationFile) WriteEvaluationRecord(model model.Model, language langu
...
+	new.Assessments = metrics.Merge(e.Assessments, nil)
+
+	return new
+}
+
+// EvaluationRecords holds all the evaluation records.
+type EvaluationRecords []*EvaluationRecord
+
+// EvaluationRecordsPerModel holds the collection of evaluation records per model.
+type EvaluationRecordsPerModel map[string]*EvaluationRecord
```
bauersimon, line 119: Once we get the model costs and names on the fly, this could even be just `map[string][]*metrics.Assessments`.

bauersimon: Similar for "by language" below.

evaluate/report/csv_test.go
```diff
@@ -154,3 +177,602 @@ func TestWriteEvaluationRecord(t *testing.T) {
...
+		EvaluationRecords EvaluationRecords
+
+		ExpectedEvaluationRecords map[string]*EvaluationRecord
+	}
+
+	validate := func(t *testing.T, tc *testCase) {
+		t.Run(tc.Name, func(t *testing.T) {
+			actualEvaluationRecords := tc.EvaluationRecords.GroupByModel()
+
+			expected, actual := maps.Clone(tc.ExpectedEvaluationRecords), maps.Clone(actualEvaluationRecords)
```
bauersimon, line 340: Why do we need to clone?

evaluate/report/csv_test.go
```diff
@@ -154,3 +177,602 @@ func TestWriteEvaluationRecord(t *testing.T) {
...
+		metrics.AssessmentKeyFilesExecuted:                      11,
+		metrics.AssessmentKeyGenerateTestsForFileCharacterCount: 12,
+		metrics.AssessmentKeyProcessingTime:                     13,
+		metrics.AssessmentKeyResponseCharacterCount:             14,
+		metrics.AssessmentKeyResponseNoError:                    15,
+		metrics.AssessmentKeyResponseNoExcess:                   16,
+		metrics.AssessmentKeyResponseWithCode:                   17,
+	}
+
+	fmt.Println(a.Score())
```
bauersimon, line 741: Println.

evaluate/report/csv_test.go
```diff
@@ -154,3 +177,602 @@ func TestWriteEvaluationRecord(t *testing.T) {
...
+	a := metrics.Assessments{
+		metrics.AssessmentKeyCoverage:                           10,
+		metrics.AssessmentKeyFilesExecuted:                      11,
+		metrics.AssessmentKeyGenerateTestsForFileCharacterCount: 12,
+		metrics.AssessmentKeyProcessingTime:                     13,
+		metrics.AssessmentKeyResponseCharacterCount:             14,
+		metrics.AssessmentKeyResponseNoError:                    15,
+		metrics.AssessmentKeyResponseNoExcess:                   16,
+		metrics.AssessmentKeyResponseWithCode:                   17,
+	}
```
bauersimon, line 739: Never used.

bauersimon (changes requested):

evaluate/report/csv.go
ruiAzevedo19: Follow up

evaluate/report/csv_test.go
ruiAzevedo19: It looks like we still need the clone otherwise the scoring is not correct. That's because in the `WriteCSVs` we are grouping the records in two different ways and we are using the original evaluation record pointers, so we need to clone them.

bauersimon: If we need something special in the tests to make it work, then we should solve that in the actual code. Cause that means we could forget to do that in the real code :see_no_evil:

ruiAzevedo19: Overlooked the comment, I thought you were talking about the `Clone` function for `EvaluationRecords`.

ruiAzevedo19: Removed the `maps.Clone` since it is not needed

bauersimon (approved):

PR #244, opened 2024-07-05
Update 0.5.0 report with results for Gemma 2 model family

bauersimon:


zimmski:
Please add rows to the overall-eval CSVs as well

PR #245, opened 2024-07-08
Structural logging

ahumenberger:


log/logger.go
```diff
@@ -1,27 +1,106 @@
...
+)
+
+type Flags int
+
+const (
+	FlagMessageOnly = 0
+	FlagDate        = 1 << iota
+	FlagTime
+	FlagStandard = FlagDate | FlagTime
+)
```
bauersimon, line 37: Docstrings. Not crazy necessary for the keys/flags themselves but at least for the flag.

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+// NeedsSpawn returns if a new log file has to be spawned.
+func (s logFileSpawner) NeedsSpawn(attributes map[AttributeKey]string) bool {
+	for _, attributeKey := range s.NeededAttributes {
+		if value := attributes[attributeKey]; value == "" {
+			return false
+		}
+	}
+
+	return true
+}
```
bauersimon, line 315: that is super clever ^^

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+		buffer = buffer[:len(buffer)-1]
+	}
+
+	var pc uintptr
+	record := slog.NewRecord(time.Now(), slog.LevelInfo, string(buffer), pc)
+
+	return originalLength, w.handler.Handle(context.Background(), record)
+}
+
+var initialLogFileSpawners = []logFileSpawner{
```
bauersimon, line 268: maybe `defaultLogFileSpawners`

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+	newHandler := h.Clone()
+
+	// Check if we need to spawn a new log file.
+	for i, spawner := range h.logFileSpawners {
+		if spawner.NeedsSpawn(h.attributes) {
+			logFilePath := spawner.FilePath(h.attributes)
+			writer, err := newLogWriter(h.writer, logFilePath)
+			if err != nil {
+				fmt.Fprintf(h.writer, "ERROR: cannot create new handler: %s\n", err.Error())
+			}
```
bauersimon, line 227: should probably return from the function then?

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+
+			logMessage := fmt.Sprintf("Spawning new log file at %s", logFilePath)
+			h.Handle(context.Background(), slog.NewRecord(time.Now(), slog.LevelInfo, logMessage, 0))
+
+			newHandler.writer = writer
+			newHandler.logFileSpawners = slices.Delete(newHandler.logFileSpawners, i, i+1) // The currently triggered log file spawner must not be part of the new handler as it would trigger again and again.
```
bauersimon, line 233: can we delete  the entry from the slice while we are iterating over it?

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+	// Collect attributes.
+	for _, attribute := range attributes {
+		h.attributes[AttributeKey(attribute.Key)] = attribute.Value.String()
+	}
+
+	newHandler := h.Clone()
+
+	// Check if we need to spawn a new log file.
+	for i, spawner := range h.logFileSpawners {
+		if spawner.NeedsSpawn(h.attributes) {
```
bauersimon, line 222: could do early exit, i.e. `if !spawner.NeedsSpawn(h.attributes) { continue }`

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+	}
+	if h.flags&FlagTime != 0 {
+		fmt.Fprint(h.writer, record.Time.Format("15:04:05"))
+		fmt.Fprint(h.writer, " ")
+	}
+
+	fmt.Fprintln(h.writer, record.Message)
+
+	return nil
+}
```
bauersimon, line 208: is not handling the attributes of the record itself... right? so we always have to pass the attributes to the `handler.WithAttrs`?

would maybe use `record.NumAttrs` to check that and throw/log an error in case we get attributes here

log/logger.go
```diff
@@ -42,34 +121,206 @@ func File(path string) (logger *Logger, loggerClose func(), err error) {
...
+
+// WithAttrs returns a new Handler whose attributes consist of both the receiver's attributes and the arguments.
+// The Handler owns the slice: it may retain, modify or discard it.
+func (h *spawningHandler) WithAttrs(attributes []slog.Attr) slog.Handler {
```
bauersimon, line 212: So the moment we create a new handler with attributes we check if there needs to be a file that needs to be created... Only thing that is not handled is what happens to the attributes when we don't have enough to start a file... cause those values are then kinda lost...

so maybe we have a slice `unusedAttributes`  that contains the attributes which are not yet required by any spawner... and then those attributes are printed with the logs... though we hopefully won't have this exact case in our setup so we might as well ignore it

evaluate/evaluate.go
```diff
@@ -89,18 +89,24 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore, totalScore uin
...
+				logger := logger.With(log.AttributeKeyLanguage, language.ID())
+
 				languageID := language.ID()
 				repositoryPath := filepath.Join(language.ID(), RepositoryPlainName)
 				temporaryRepository := temporaryRepositories[repositoryPath]
 
+				logger = logger.With(log.AttributeKeyRepository, temporaryRepository.Name())
 				for _, model := range ctx.Models {
 					modelID := model.ID()
+					logger := logger.With(log.AttributeKeyModel, modelID)
```
bauersimon, line 109: that is pretty sweet

log/logger_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/zimmski/osutil"
+	"github.com/zimmski/osutil/bytesutil"
+)
+
+func TestLoggerWith(t *testing.T) {
```
bauersimon, line 16: awesoooome

tiny nit: just for fun, I'd log something to the final logger objects you create... cause you check now that the files are created but just to make sure stuff ends up there

bauersimon (changes requested):

log/logger.go
ahumenberger: `// Flags defines how log messages should be printed.`

log/logger.go
ahumenberger: we iterate over `h.logFileSpawners` and delete from the clone `newHandler.logFileSpawners`

log/logger.go
ahumenberger: Not sure. We still have a valid logger where we can print the error message to. returning a nil handler here seems  to be not intended.

log/logger.go
ahumenberger: hmm. would not do that I think. 

if we want to log query responses we probably do something like `log.Info(response, AttributeKeyArtifact, ArtifactTypeQueryResponse)`. The attribute is here only valid for the log message, we do not need to create a separate logger for that.

log/logger.go
ahumenberger: Yes they are "lost". Right now every log message has still all relevant information in the message itself. E.g 

```
2024/07/08 09:39:00 Evaluated model "symflower/symbolic-execution" on task "write-tests" using language "java" and repository "java/plain": encountered 0 problems: []
```
model, repository, language and task is still contained. but at one point we should switch to something like

```
2024/07/08 09:39:00 [model=symflower/symbolic-execution][language=java][task=write-tests][repository=java/plain] Evaluation done. Encountered 0 problems: []
```

And e.g. the log file for a specific task then contains the same log message but without the "task" attribute printed.

log/logger.go
bauersimon: the writer will be `nil` though, right? so maybe we just reset the writer to the one we were using

log/logger.go
bauersimon: I guess calls like `log.Infof` with specific attributes are anyways not possible yet, so the attributes can never show up here.

log/logger.go
ahumenberger: the new writer is nil, but we are logging to it's parent `h.writer`

bauersimon: below we set `newHandler.writer = writer` and return that as new handler... so when that handler is used and tries to log to its own writer :boom:

bauersimon: I think in this case we should do `continue` and ignore the current spawner

ahumenberger: ah sure. the `continue` is missing. sorry.

evaluate/evaluate_test.go
```diff
@@ -219,7 +219,7 @@ func TestEvaluate(t *testing.T) {
 			ExpectedResultFiles: map[string]func(t *testing.T, filePath string, data string){
 				"evaluation.log": nil,
 				filepath.Join(string(evaluatetask.IdentifierWriteTests), mockedModel.ID(), "golang", "golang", "plain.log"): nil,
-				filepath.Join("evaluation.csv"): nil,
+				"evaluation.csv": nil,
```
bauersimon, line 222: nit: `refactor,`

bauersimon:
@ahumenberger gave up trying to fix this, was even able to verify that the GC properly closes the files on Linux. skipping the test on windows now...

```sh
go test ./log/... -run TestLoggerWith -count 1 -v
=== RUN   TestLoggerWith
=== RUN   TestLoggerWith/New_log_file_for_Result_path
GC
=== RUN   TestLoggerWith/New_log_file_for_Language-Model-Repository-Task
GC
GC
=== RUN   TestLoggerWith/New_log_file_for_two_tasks
GC
GC
GC
=== RUN   TestLoggerWith/New_log_file_for_two_repositories
GC
GC
GC
--- PASS: TestLoggerWith (0.01s)
    --- PASS: TestLoggerWith/New_log_file_for_Result_path (0.01s)
    --- PASS: TestLoggerWith/New_log_file_for_Language-Model-Repository-Task (0.00s)
    --- PASS: TestLoggerWith/New_log_file_for_two_tasks (0.00s)
    --- PASS: TestLoggerWith/New_log_file_for_two_repositories (0.00s)
PASS
ok      github.com/symflower/eval-dev-quality/log       0.020s
```

ahumenberger:
> @ahumenberger gave up trying to fix this, was even able to verify that the GC properly closes the files on Linux. skipping the test on windows now...


Solved it differently with

```go
func addOpenLogFile(file *os.File) {
	openLogFilesMutex.Lock()
	defer openLogFilesMutex.Unlock()

	openLogFiles = append(openLogFiles, file)
}

// CloseOpenLogFiles closes the files that were opened by some logger.
func CloseOpenLogFiles() {
	openLogFilesMutex.Lock()
	defer openLogFilesMutex.Unlock()

	for _, logFile := range openLogFiles {
		logFile.Close()
	}

	openLogFiles = nil
}
```

bauersimon (approved):

PR #249, opened 2024-07-09
Use "main" image if no image was specified

Munsio:


Munsio:
@bauersimon this is like a quick workaround to use the latest available image from the registry when using a newer version checked out from github.

The real solution needs https://github.com/symflower/eval-dev-quality/issues/207 to go in first and then build up from there.

PR #251, opened 2024-07-09
Revert the docker image to be a root user container

Munsio:


Munsio:
As this is only temporary I commited the comments too for easier reverting.

PR #255, opened 2024-07-10
Add the commit revision to binary/reports/docker

Munsio:


Munsio:
@ruiAzevedo19 please review

Testing the revision with the binary and a golang test itself would mean to run system tests which we currently do not have. Therefore the normal test with the check for `development` is the only one.

The docker image one can be reviewed here: https://github.com/symflower/eval-dev-quality/actions/runs/9875489063/job/27272302420?pr=255#step:6:41
Please note that the Branch Commit hash and the PR commit hash are different (unfortunately) But  we are building on Branches now too which you can check with: https://github.com/symflower/eval-dev-quality/actions/runs/9875489063/job/27272302420#step:6:41 This one has to be the same as the latest commit in the branch itself.

.github/workflows/docker.yml
```diff
@@ -34,3 +34,5 @@ jobs:
           platforms: linux/amd64
           tags: ${{ steps.meta.outputs.tags }}
           push: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
ruiAzevedo19, line 38: Can you explain me this please?

.github/workflows/branch.yml
```diff
@@ -0,0 +1,15 @@
+name: Branch CI
+
+on:
+  push:
+    branches-ignore:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
ruiAzevedo19, line 10: The group here creates an ID for the current CI run right?

cmd/eval-dev-quality/cmd/version.go
```diff
@@ -0,0 +1,31 @@
...
+
+// SetLogger sets the logger of the command.
+func (command *Version) SetLogger(logger *log.Logger) {
+	command.logger = logger
+}
+
+// Execute executes the command.
+func (command *Version) Execute(args []string) (err error) {
+	revision := evaluate.Revision
+	if evaluate.Revision == "" {
```
ruiAzevedo19, line 24: Maybe we can use `if revision == "" { ... }`

.github/workflows/branch.yml
Munsio: Yes. With this line we tell Github Actions to only have the latest CI running and abort the previous one. This is especially helpful if you push multiple times in a short time.

.github/workflows/docker.yml
Munsio: This is to tell the Docker build step to push/fetch data from the Github Cache so building the container is faster on subsequent runs.

ruiAzevedo19: Thanks

ruiAzevedo19:
@bauersimon Please, do a second review

.github/workflows/branch.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  docker:
+    name: 'Run docker image building'
+    uses: ./.github/workflows/docker.yml
```
bauersimon, line 15: what is with the `ci.yml` workflow? it also builds the docker image right? do we have it twice now?

bauersimon (commented):
awesome - tried it locally. works great

.github/workflows/branch.yml
bauersimon: maybe we should drop the PR building then since we do the branch anyways?

bauersimon: well... we can always remove that afterwards... better to have too many than too less ^^

bauersimon (approved):

Issue #205, opened 2024-06-20
Tool/command to combine multiple evaluations into one


bauersimon:
If we run them on multiple machines, we want to easily combine them.

### TODO

**1st iteration**
- [x] Remove all CSVs but `evaluation.csv`
  - [x] Remove the `models-summed.csv` and `<language>-summed.csv` files 
  - [x] Remove all the evaluation CSV parsing logic
  - [x] Remove all occurrences of model's cost and human-readable name
  - https://github.com/symflower/eval-dev-quality/pull/256
- [x] Let the  "report" command also generate the Markdown report
  - https://github.com/symflower/eval-dev-quality/pull/258
- Reason: we want to move to a more flexible way of handling the evaluation data -> SQL

**2nd iteration**
- [x] Create a new command `eval-dev-quality report`
```golang
type Report struct {
     EvaluationPaths []string // holds  "evaluation.csv" file paths
     ResultPath string 
}
``` 

- [x] Create an `evaluation.csv` file in `ResultPath` that will hold the combined evaluation records
- [x] Loop through the `EvaluationPaths` and find all `evaluation.csv`
  - [x] Append the CSV records to the overall `evaluation.csv`

bauersimon:
Current workarounds:

- `cat docs/reports/<version>/*/models-summed.csv | sort | uniq` for overall scores
- `cat docs/reports/<version>/*/evaluation.csv | sort | uniq` for repository-specific scores

bauersimon:
- I think `eval-dev-quality report` is great
- if we want to obtain i.e. the real names and costs of the models in the reporting tool we might need to share some parts of the command arguments (i.e. the provider tokens)... though I think openrouter allows to query information without a token so we should be good for now
- I think @zimmski mentioned he wanted to use `postgres` now ... not sure how that would work though cause it seems you would need a running DB for that in the background?

PR #256, opened 2024-07-10
Remove the summed CSVs and model's cost and human-readable name, since they will be handled afterwards with tooling 

ruiAzevedo19:
Part of #205


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -413,22 +412,7 @@ func (command *Evaluate) evaluateLocal(evaluationContext *evaluate.Context) (err
 	}
 
 	assessments, totalScore := evaluate.Evaluate(evaluationContext)
-
 	assessmentsPerModel := assessments.CollapseByModel()
-	if err := (report.Markdown{
```
bauersimon, line 419: Please move this commit into a separate PR. Not 100% sure if we want to scrap that yet. Good to have. But will not merge that yet.

bauersimon (changes requested):

bauersimon (approved):

Issue #257, opened 2024-07-11
Change all prompts to enforce code fences


bauersimon:
- [x] "The response must contain only the source code in a fenced code block and nothing else."
  - [x] Write test
  - [x] Transpile
  - [x] Code repair 
- [x] close https://github.com/symflower/eval-dev-quality/issues/43

PR #259, opened 2024-07-11
Change the "code-repair" and "write-tests" task prompt to enforce the generated code to be inside code fences, to ensure we can extract the code from the LLMs response

ruiAzevedo19:
Part of #257


zimmski:
LGMT

@ruiAzevedo19 please make sure to include the same change in the transpilation task.

PR #260, opened 2024-07-11
Testing docker runtime

Munsio:


Makefile
```diff
@@ -47,13 +47,16 @@ install: # [<Go package] - # Build and install everything, or only the specified
 	go install -v -ldflags="$(GO_LDFLAGS)" $(PACKAGE)
 .PHONY: install
 
-install-all: install install-tools-testing # Install everything for and of this repository.
+install-all: install install-tools install-tools-testing # Install everything for and of this repository.
 .PHONY: install-all
 
+install-tools: # Install tools that are required for running the evaluation.
+	eval-dev-quality install-tools $(if $(ARGS), --install-tools-path $(word 1,$(ARGS)))
+.PHONY: install-tools
```
bauersimon, line 55: wondering if we change the tools installation path then we must tell the eval binary about that path in every evaluation, right?

Dockerfile
```diff
@@ -1,40 +1,54 @@
...
 
-# Setup the evaluation
-
-RUN make install-all
-ENV PATH="${PATH}:/home/ubuntu/.eval-dev-quality/bin"
+# Install the binary
+COPY --from=builder --chown=ubuntu:ubuntu /app/eval-dev-quality /app/.eval-dev-quality/bin/
+ENV PATH="${PATH}:/app/.eval-dev-quality/bin"
+RUN make install-tools-testing
+RUN make install-tools /app/.eval-dev-quality/bin
```
bauersimon, line 54: If we change the tool installation path during the build, then we have to specify this tools path during every evaluation, no?

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -501,6 +499,68 @@ func (command *Evaluate) evaluateDocker(ctx *evaluate.Context) (err error) {
...
+					"volume-fetch",
+				},
+			})
+			if deferErr != nil {
+				err = errors.Join(err, pkgerrors.WithMessage(pkgerrors.WithStack(deferErr), output))
+			}
+		}
+
+		// Copy data from volume to filesystem.
+		output, err = util.CommandWithResult(context.Background(), command.logger, &util.Command{
```
bauersimon, line 535: err unhandled

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -501,6 +499,68 @@ func (command *Evaluate) evaluateDocker(ctx *evaluate.Context) (err error) {
...
+			})
+			if deferErr != nil {
+				err = errors.Join(err, pkgerrors.WithMessage(pkgerrors.WithStack(deferErr), output))
+			}
+		}
+
+		// Defer the cleanup in the right order, the container needs to be removed before removing the volume.
+		defer volumeCleanup()
+		defer containerCleanup()
+	}
```
bauersimon, line 562: would move the cleanup defers up to where the conainter or data volume is created to ensure 100% it is removed when intermediate errors happen

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -501,6 +499,68 @@ func (command *Evaluate) evaluateDocker(ctx *evaluate.Context) (err error) {
 	}
 	parallel.Wait()
 
+	// Copy data from volume back.
+	{
```
bauersimon, line 503: nit: would be nice to have this volume copying done in a helper method with some tests

can be follow-up

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -676,6 +679,249 @@ func TestEvaluateExecute(t *testing.T) {
...
+		} else if _, err := exec.LookPath("docker"); err != nil {
+			t.Skip("Docker runtime not found")
+		}
+
+		// TODO: Make the docker runtime run on both local and CI
+		if os.Getenv("GITHUB_ACTIONS") != "true" {
+			t.Skip("We are currently only supporting the Github CI")
+		}
+
+		// Prerequisite
```
bauersimon, line 695: docstring that states we are are pulling the image that was just built for this branch to avoid building it again

cmd/eval-dev-quality/cmd/evaluate.go
Munsio: The docker tests are already testing the behavior that everything is copied back from the volumes to the temporary test result path.

Dockerfile
Munsio: Do we? As it is inside the `PATH` it shouldn't matter because we can resolve it inside the command with a lookup.

Dockerfile
```diff
@@ -1,40 +1,54 @@
...
 
 # Install Go
 RUN wget https://go.dev/dl/go1.21.5.linux-amd64.tar.gz && \
-	tar -xf go1.21.5.linux-amd64.tar.gz -C ~/.eval-dev-quality/ && \
+	tar -xf go1.21.5.linux-amd64.tar.gz -C /app/.eval-dev-quality/ && \
 	rm go1.21.5.linux-amd64.tar.gz
-ENV PATH="${PATH}:/home/ubuntu/.eval-dev-quality/go/bin"
+ENV PATH="${PATH}:/app/.eval-dev-quality/go/bin"
+ENV GOROOT="/app/.eval-dev-quality/go"
 ENV PATH="${PATH}:/home/ubuntu/go/bin"
```
ruiAzevedo19, line 48: Is there any reason why `home/ubuntu/go/bin` is still added to the path and not only `app/.eval-dev-quality/go/bin`?

Munsio: Yes because the GOPATH changes unfortunatelly depending on the User, it is not ideal but currently the only possible workaround.

ruiAzevedo19:
@Munsio Done with the review

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -701,6 +704,266 @@ func TestEvaluateExecute(t *testing.T) {
...
+			_, logger := log.Buffer()
+			branch, err := util.CommandWithResult(context.Background(), logger, &util.Command{
+				Command: []string{
+					"git",
+					"rev-parse",
+					"--short",
+					"HEAD",
+				},
+			})
+			assert.NoError(t, err)
```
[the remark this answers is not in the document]
Munsio: Good catch. Fixed

Issue #230, opened 2024-07-01
Make the Knapsack.java case easier to solve for models


zimmski:
```
[`Knapsack.java`](https://github.com/symflower/eval-dev-quality/blob/main/testdata/java/light/src/main/java/com/eval/Knapsack.java) contains a nested class `Item`. Due to an oversight on our side we did not make the class `static` which means `Item` needs to be initialized with `new Knapsack().new Item()`. We were surprised to see that this caused troubles for almost all LLMs which were trying to initialize with `new Knapsack.Item()`.
```

Since these examples are not about language features but about how well LLMs are doing at logic and general use of a language, please fix the example so it is easier to solve.

PR #262, opened 2024-07-11
Make the Java Knapsack inner class static, so it is easier for LLMs to solve 

ruiAzevedo19:
Part of #230

### DIFF
--- a/testdata/java/light/src/main/java/com/eval/Knapsack.java
+++ b/testdata/java/light/src/main/java/com/eval/Knapsack.java
@@ -26,7 +26,7 @@ public class Knapsack {
     return knapsack[items.length][maximumWeight];
   }
 
-  class Item {
+  static class Item {
     int weight;
     int value;
   }


testdata/java/light/src/main/java/com/eval/Knapsack.java
```diff
@@ -1,33 +1,33 @@
...
       }
     }
-    return knapsack[items.length][maximumWeight];
-  }
 
-  class Item {
-    int weight;
-    int value;
+    return knapsack[numberOfItems][maximumWeight];
   }
```
bauersimon, line 32: I think it is enough to make this `public static class Item`. Cause that was causing the big confusion. Also now you would need to adapt the tests :see_no_evil:

testdata/golang/light/knapsack.go
```diff
@@ -1,34 +1,37 @@
...
-import "math"
+import (
+	"errors"
+	"math"
+)
 
-type Item struct {
-	weight int
-	value  int
-}
```
bauersimon, line 8: I think the Go example was fine. Did not cause any mistakes. LLMs should be able to deal with structs as arguments...

testdata/java/light/src/main/java/com/eval/Knapsack.java
ruiAzevedo19: This is the light repository, there is no predefined tests, and we don't use the knapsack example in other repositories

Issue #263, opened 2024-07-12
Check if all testdata repositories are well-formed just once, and not in every task run


ruiAzevedo19:
Goal: check if the testdata repositories are well-formed just once and not in every run.

Currently, we check if a task repository is well-formed when we run the task itself. Since we can have multiple runs, we are checking the same thing over and over again. There needs to be an extra step before running all the tasks where we validate the testdata repositories.

#### TODO

- [x] add a new method to the repository interface `Validate() error` 
- [x] each task file contains a validation function that knows how to validate the repository 
  - [x] write-tests task
  - [x] code-repair task
  - [x] transpilation task 
- [x] the repository validation needs to happen in 2 places in `evaluate/evaluate.go:Evaluate`
  - [x] when we run the evaluation for the `plain` repository (after calculating the `temporaryRepositories` and before the evaluation loop)
  - [x] when we run the evaluation for all repositories  (after calculating the `temporaryRepositories` and before the evaluation loop)

bauersimon:
Would basically just extend the 

```go
type Repository interface { 
	Validate() error
...
```

and that way you can just call that function in the two locations you mentioned above right after creating the repository

and internally you will probably do a switch based on the task identifiers that the repository has... and do one special validation function per task... would probably even leave the validation logic itself where the tasks are (i.e. `task-write-test.go`)  that way the validation for a task is still where the task happens

bauersimon:
so basically i would not do a function that knows how to validate all types of repository... I think it is still good that this logic is separated... but the way that the repository validation is invoked should just be via the `Repository` interface and only once when the repository is created

PR #265, opened 2024-07-12
Check if the testdata repository is valid before running the evaluation, so it is checked just once

ruiAzevedo19:
Part of #263

### DIFF
--- a/evaluate/task/task-code-repair_test.go
+++ b/evaluate/task/task-code-repair_test.go
@@ -1,6 +1,8 @@
 package task
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


evaluate/task/task-write-test_test.go
```diff
@@ -216,3 +217,114 @@ func TestTaskWriteTestsRun(t *testing.T) {
...
+				tc.Before(repositoryPath)
+			}
+
+			actualErr := validateWriteTestsRepository(logger, repositoryPath, tc.Language)
+			if len(tc.ExpectedErr) > 0 {
+				require.ErrorContains(t, actualErr, tc.ExpectedErr)
+			} else {
+				require.NoError(t, actualErr)
+			}
+		})
```
bauersimon, line 256: I think at this point it would be cool to refactor this validation (and the `testCase`) out and reuse across the different tasks.

evaluate/task/task-write-test_test.go
```diff
@@ -216,3 +217,114 @@ func TestTaskWriteTestsRun(t *testing.T) {
...
+				Name: "Repository with test files",
+
+				Before: func(repositoryPath string) {
+					// Create test files to trigger an error.
+					file, err := os.Create(filepath.Join(repositoryPath, "foo_test.go"))
+					require.NoError(t, err)
+					defer file.Close()
+					file, err = os.Create(filepath.Join(repositoryPath, "bar_test.go"))
+					require.NoError(t, err)
+					defer file.Close()
```
bauersimon, line 279: Same suggestions as before. Please only create as many files as you need to trigger an error. And don't use `foo` and `bar` but what the entity actually is: `somefile`, `somepackage`, etc.

bauersimon (changes requested):

task/task.go
```diff
@@ -54,6 +54,9 @@ type Repository interface {
 	// SupportedTasks returns the list of task identifiers the repository supports.
 	SupportedTasks() (tasks []Identifier)
 
+	// Validate checks it the repository is well-formed.
+	Validate(logger *log.Logger, language language.Language) (err error)
+
```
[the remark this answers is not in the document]
ruiAzevedo19: Follow-up #266

evaluate/task/task-code-repair_test.go
```diff
@@ -267,3 +269,206 @@ func TestTaskCodeRepairRun(t *testing.T) {
 		}
 	})
 }
+
+func TestValidateCodeRepairRepository(t *testing.T) {
+	validate := func(t *testing.T, tc *tasktesting.TestCaseValidateRepository) {
+		t.Run(tc.Name, func(t *testing.T) {
+			tc.Validate(t, validateCodeRepairRepository)
+		})
+	}
```
bauersimon, line 278: Please move the `t.Run` into the actual validation function. That way you don't need to define your own here.

Issue #167, opened 2024-06-06
Fixed timeouts for `symflower unit-tests` and `symflower test`


Munsio:
During the run of the light repository with the ollama models I encountered a problem where `symfower test` was running for ~1H (stuck). In order to avoid this we need to set sane timeouts for all those execution calls.

- [x] `symflower test` #188 
- [x] `symflower unit-tests`#267

### Notes:
- Maven downloads may take also a while, should we add a "prefetch" state for dependencies so we can retry in the case of an error. (We know that maven is sometimes flaky)
- I think a sane execution time limit for running tests would be 5-10 minutes. If we are going for a prefetch stage than it should be on the lower end.
- We should probably also use the timeout error as metric.

Munsio:
language/java/language.go:96
language/golang/language.go:100
model/symflower/symflower.go:42

I guess those are the parts where we need to apply the timeout.

PR #267, opened 2024-07-15
fix, Define a timeout for the "symflower unit-tests" command, so ensure the execution does not take too much time

ruiAzevedo19:
Part of #167

### DIFF
--- a/model/symflower/symflower.go
+++ b/model/symflower/symflower.go
@@ -16,12 +16,27 @@ import (
 	"github.com/symflower/eval-dev-quality/util"
 )
 
+// defaultSymbolicExecutionTimeout defines the default symbolic execution timeout.
+var defaultSymbolicExecutionTimeout = time.Duration(10 * time.Minute)
+
 // Model holds a Symflower model using the locally installed CLI.
-type Model struct{}
+type Model struct {
+	// symbolicExecutionTimeout defines the symbolic execution timeout.
+	symbolicExecutionTimeout time.Duration
+}
 
 // NewModel returns a Symflower model.
 func NewModel() (model *Model) {
-	return &Model{}
+	return &Model{
+		symbolicExecutionTimeout: defaultSymbolicExecutionTimeout,
+	}
+}
+
+// NewModelWithTimeout returns a Symflower model with a given timeout.
+func NewModelWithTimeout(timeout time.Duration) (model *Model) {
+	return &Model{
+		symbolicExecutionTimeout: timeout,
+	}
 }
 
 var _ model.Model = (*Model)(nil)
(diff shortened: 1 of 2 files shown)


model/symflower/symflower_test.go
```diff
@@ -257,3 +260,38 @@ func TestCountCharactersOfGeneratedFiles(t *testing.T) {
...
+
+		Directory: repositoryPath,
+	})
+	duration := time.Since(start)
+
+	if osutil.IsWindows() {
+		assert.ErrorContains(t, err, context.DeadlineExceeded.Error())
+	} else {
+		assert.ErrorContains(t, err, "signal: killed")
+	}
```
ahumenberger, line 295: Does `assert.ErrorIs` not work here?

model/symflower/symflower.go
```diff
@@ -35,9 +36,12 @@ var _ model.CapabilityWriteTests = (*Model)(nil)
 
 // generateTestsForFile generates test files for the given implementation file in a repository.
 func (m *Model) WriteTests(ctx model.Context) (assessment metrics.Assessments, err error) {
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), language.DefaultExecutionTimeout)
+	defer cancel()
+
```
ahumenberger, line 41: I guess this should somehow be added to the assessment that this ran into a timeout.

@bauersimon

model/symflower/symflower.go
```diff
@@ -35,9 +36,12 @@ var _ model.CapabilityWriteTests = (*Model)(nil)
 
 // generateTestsForFile generates test files for the given implementation file in a repository.
 func (m *Model) WriteTests(ctx model.Context) (assessment metrics.Assessments, err error) {
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), language.DefaultExecutionTimeout)
```
bauersimon, line 39: hmmmmmmmm not really sure about using the language timeout... cause symflower is a separate model. I.e. we might want to give more time to the SE than to the test execution.

How about adding a new value `SymbolicExecutionTimeout` to the `tools` package and a `Model.timeout` field to this model. That field can then be initialized with `tools.SymbolicExecutionTimeout` but overwritten for the tests.

model/symflower/symflower_test.go
```diff
@@ -257,3 +260,38 @@ func TestCountCharactersOfGeneratedFiles(t *testing.T) {
 		ExpectedCount: 169,
 	})
 }
+
+func TestSymflowerUnitTestsWithTimeout(t *testing.T) {
```
bauersimon, line 264: Not a big fan of having some functionality and then rebuilding the same functionality for a tests.

model/symflower/symflower_test.go
bauersimon: I guess it might. But I remember that the signal errors are actually quite nasty structurally so a string is way simpler.

model/symflower/symflower.go
bauersimon: But if we adapt the tests we will see for sure.

bauersimon (dismissed):

tools/symflower.go
```diff
@@ -9,13 +9,17 @@ import (
...
 
 	pkgerrors "github.com/pkg/errors"
 	"github.com/symflower/eval-dev-quality/log"
 	"github.com/symflower/eval-dev-quality/util"
 	"github.com/zimmski/osutil"
 )
 
+// SymbolicExecutionTimeout defines the symbolic execution timeout.
+var SymbolicExecutionTimeout = time.Duration(10 * time.Minute)
+
```
ahumenberger, line 22: Why not define that in `model/symflower`?

ahumenberger (changes requested):

ahumenberger (approved):

Issue #225, opened 2024-06-27
Do not start the ollama server if not needed


Munsio:
During my many runs for the docker version it always came across that we are starting the ollama server as soon as we run `evaluate` which is kinda annoying as I am using `symflower/symbolic-execution` or other models and those do not need the ollama server running in the background.

PR #269, opened 2024-07-15
Shut down the Ollama service if there are no Ollama models to evaluate, to avoid unnecessary processes being opened

ruiAzevedo19:
Part of #225

### DIFF
--- a/cmd/eval-dev-quality/cmd/evaluate_test.go
+++ b/cmd/eval-dev-quality/cmd/evaluate_test.go
@@ -525,6 +525,31 @@ func TestEvaluateExecute(t *testing.T) {
 						filepath.Join("result-directory", "README.md"): nil,
 						filepath.Join("result-directory", string(evaluatetask.IdentifierWriteTests), "ollama_"+log.CleanModelNameForFileSystem(providertesting.OllamaTestModel), "golang", "golang", "plain.log"): nil,
 					},
+					ExpectedOutputValidate: func(t *testing.T, output, resultPath string) {
+						assert.Contains(t, output, `Starting services for provider "ollama"`)
+					},
+				})
+			}
+			{
+				validate(t, &testCase{
+					Name: "Ollama services are not started",
+
+					Arguments: []string{
+						"--language", "golang",
+						"--model", "symflower/symbolic-execution",
+						"--repository", filepath.Join("golang", "plain"),
+					},
+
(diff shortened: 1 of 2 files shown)


Munsio:
Out of curiosity I checked this PR and I have to say instead of shutting down the ollama server we shouldn't even start it if it is not needed.

Also @ruiAzevedo19 it is Issue #225 not #255

bauersimon:
Yeah I think we should maybe reorder that the providers are parsed from the CLI first... and then we don't even start the ones that are not needed :thinking:

bauersimon:
Well if there is no model that starts with `ollama/` we will never have an ollama model.

ruiAzevedo19:
> Well if there is no model that starts with `ollama/` we will never have an ollama model.

We only add the `ollama/` prefix after fetching the Ollama models

Munsio:
> > Well if there is no model that starts with `ollama/` we will never have an ollama model.
> 
> We only add the `ollama/` prefix after fetching the Ollama models

And that needs to be swapped. Parse the CLI first and only parse models of providers which are needed. Then do the validation.

Munsio (approved):

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -525,6 +525,31 @@ func TestEvaluateExecute(t *testing.T) {
...
+
+					ExpectedResultFiles: map[string]func(t *testing.T, filePath string, data string){
+						filepath.Join("result-directory", "categories.svg"): nil,
+						filepath.Join("result-directory", "evaluation.csv"): nil,
+						filepath.Join("result-directory", "evaluation.log"): nil,
+						filepath.Join("result-directory", "README.md"):      nil,
+						filepath.Join("result-directory", string(evaluatetask.IdentifierWriteTests), "symflower_symbolic-execution", "golang", "golang", "plain.log"): nil,
+					},
+					ExpectedOutputValidate: func(t *testing.T, output, resultPath string) {
+						assert.NotContains(t, output, "Starting services for provider \"ollama\"")
```
Munsio, line 551: Sry for the nitpick but please change this to: 
```
assert.NotContains(t, output, `Starting services for provider "ollama"`)
```

The other one too.

Munsio (changes requested):

Munsio (approved):

PR #271, opened 2024-07-17
Introduce the "report" command to combine multiple evaluations into a single file

ruiAzevedo19:
Part of #205

### DIFF
--- a/cmd/eval-dev-quality/cmd/report.go
+++ b/cmd/eval-dev-quality/cmd/report.go
@@ -1,0 +1,108 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+
+	pkgerrors "github.com/pkg/errors"
+	"github.com/zimmski/osutil"
+	"golang.org/x/exp/maps"
+
+	"github.com/symflower/eval-dev-quality/evaluate/report"
+	"github.com/symflower/eval-dev-quality/log"
+)
+
+// Report holds the "report" command.
+type Report struct {
+	// EvaluationPaths holds the list of file paths where the results of previous evaluations are stored.
+	EvaluationPaths []string `long:"evaluation-path" description:"File path for an evaluation CSV file where the results of a previous evaluation are stored."`
+	// ResultPath holds the directory path where the overall results should be written to.
+	ResultPath string `long:"result-path" description:"Directory path were the combined results are written to."`
+
+	// logger holds the logger of the command.
+	logger *log.Logger
+}
+
+var _ SetLogger = (*Evaluate)(nil)
+
+// SetLogger sets the logger of the command.
+func (command *Report) SetLogger(logger *log.Logger) {
+	command.logger = logger
+}
+
+// Execute executes the command.
+func (command *Report) Execute(args []string) (err error) {
+	// Create the result path directory and check if there is already an evaluation CSV file in there.
+	var evaluationCSVFile *os.File
+	if err = osutil.MkdirAll(filepath.Dir(command.ResultPath)); err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+	if _, err := os.Stat(filepath.Join(command.ResultPath, "evaluation.csv")); err != nil {
+		if os.IsNotExist(err) {
+			evaluationCSVFile, err = os.Create(filepath.Join(command.ResultPath, "evaluation.csv"))
+			if err != nil {
+				command.logger.Panicf("ERROR: %s", err)
+			}
+			defer evaluationCSVFile.Close()
+		} else {
+			command.logger.Panicf("ERROR: %s", err)
+		}
+	} else {
+		command.logger.Panicf("ERROR: an evaluation CSV file already exists in %s", command.ResultPath)
+	}
+
+	// Collect all evaluation CSV file paths.
+	allEvaluationPaths := map[string]bool{}
+	for _, evaluationPath := range command.EvaluationPaths {
+		evaluationFilePaths, err := pathsFromGlobPattern(evaluationPath)
+		if err != nil {
+			command.logger.Panicf("ERROR: %s", err)
+		}
+		for _, evaluationPath := range evaluationFilePaths {
+			allEvaluationPaths[evaluationPath] = true
+		}
+	}
+
+	// Collect all records from the evaluation CSV files.
+	records, err := report.RecordsFromEvaluationCSVFiles(maps.Keys(allEvaluationPaths))
+	if err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	} else if len(records) == 0 {
+		command.logger.Printf("no evaluation records found in %+v", command.EvaluationPaths)
+
+		return nil
+	}
+	report.SortEvaluationRecords(records)
+
+	// Write all records into a single evaluation CSV file.
+	evaluationFile, err := report.NewEvaluationFile(evaluationCSVFile)
+	if err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+	if err := evaluationFile.WriteLines(records); err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


cmd/eval-dev-quality/cmd/report.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/symflower/eval-dev-quality/evaluate/report"
+	"github.com/symflower/eval-dev-quality/log"
+)
+
+// Report holds the "report" command.
+type Report struct {
+	// EvaluationPaths holds the list of file paths where the results of previous evaluations are stored.
+	EvaluationPaths []string `long:"evaluation-path" description:"File path for an evaluation CSV file where the results of a previous evaluation are stored."`
+	// ResultPath holds the directory path where the overall results should be written to.
+	ResultPath string `long:"result-path" description:"Directory path were the overall results are written to."`
```
Munsio, line 21: Would change to `where the combined results are written to`

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -0,0 +1,109 @@
...
+var _ SetLogger = (*Evaluate)(nil)
+
+// SetLogger sets the logger of the command.
+func (command *Report) SetLogger(logger *log.Logger) {
+	command.logger = logger
+}
+
+// Execute executes the command.
+func (command *Report) Execute(args []string) (err error) {
+	evaluationCSVFilePath, err := os.OpenFile(filepath.Join(command.ResultPath, "evaluation.csv"), os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0755)
```
Munsio, line 36: We do know that the file does not exist but we do NOT know if the directory already exists. Would change this to something like this:  `evaluate/report/markdown.go:193`
```go
if err = os.MkdirAll(filepath.Dir(path), 0755); err != nil {
		return pkgerrors.WithStack(err)
	}
	file, err := os.Create(path)
	if err != nil {
		return pkgerrors.WithStack(err)
	}
```

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+	defer evaluationCSVFilePath.Close()
+
+	evaluationCSVWriter := bufio.NewWriter(evaluationCSVFilePath)
```
Munsio, line 42: Would move the Writer below the gathering of records, and check if there are any records to begin with.

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -0,0 +1,109 @@
...
+		records, err := recordsFromGlobPattern(evaluationPath)
+		if err != nil {
+			command.logger.Panicf("ERROR: %s", err)
+		}
+		allRecords = append(allRecords, records...)
+	}
+
+	sort.Strings(allRecords)
+	for _, record := range allRecords {
+		if _, err = evaluationCSVWriter.WriteString(record + "\n"); err != nil {
```
Munsio, line 58: Add this one to `evaluate/report/csv.go` and create a new Method where you can write a row based on a string slice. With that change it is testable and can be used by the existing parts and your new one.

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -0,0 +1,109 @@
...
+func recordsFromGlobPattern(evaluationGlobPattern string) (records []string, err error) {
+	evaluationCSVFilePaths, err := filepath.Glob(evaluationGlobPattern)
+	if err != nil {
+		return nil, pkgerrors.WithStack(err)
+	} else if len(evaluationCSVFilePaths) == 0 {
+		return nil, pkgerrors.Errorf("no files matched the pattern %q", evaluationGlobPattern)
+	}
+
+	for _, evaluationCSVFilePath := range evaluationCSVFilePaths {
+		if filepath.Base(evaluationCSVFilePath) != "evaluation.csv" {
```
Munsio, line 77: Move the check for the GlobPattern to the top and do an early exit. With a message like `the path needs to end with `evaluation.csv`

Munsio (changes requested):

cmd/eval-dev-quality/cmd/report_test.go
```diff
@@ -0,0 +1,377 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/symflower/eval-dev-quality/log"
+	"github.com/zimmski/osutil/bytesutil"
+)
+
+var claudeEvaluationCSVFileContent = bytesutil.StringTrimIndentations(`
+	model-id,language,repository,task,score,coverage,files-executed,files-executed-maximum-reachable,generate-tests-for-file-character-count,processing-time,response-character-count,response-no-error,response-no-excess,response-with-code
```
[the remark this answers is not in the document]
ruiAzevedo19: Can be a follow-up?

cmd/eval-dev-quality/cmd/report_test.go
```diff
@@ -0,0 +1,377 @@
...
+			}
+		},
+
+		ExpectedEvaluationCSVFileContent: claudeEvaluationCSVFileContent,
+	})
+	validate(t, &testCase{
+		Name: "Multiple files",
+
+		Before: func(t *testing.T, logger *log.Logger, workingDirectory string) {
+			claudeEvaluationCSVFilePath := filepath.Join(workingDirectory, "docs", "v5", "claude")
```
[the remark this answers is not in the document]
Munsio: Rly it would be faster to do it right now, not?

evaluate/report/csv.go
```diff
@@ -57,7 +61,43 @@ func (e *EvaluationFile) WriteEvaluationRecord(model model.Model, language langu
 	return nil
 }
 
+// WriteEvaluationRawRecords takes a slice of raw records and writes them into the evaluation file.
+func (e *EvaluationFile) WriteEvaluationRawRecords(rawRecords []string) (err error) {
```
Munsio, line 65: Would rename to `WriteLines` as this one is a generic method which simply writes string lines into the CSV file.

Munsio: Please also refactor the function `WriteEvaluationRecord` to use the new generic way of writing a line.

Munsio (changes requested):

evaluate/report/csv.go
```diff
@@ -57,7 +61,43 @@ func (e *EvaluationFile) WriteEvaluationRecord(model model.Model, language langu
...
+	if len(rawRecords) == 0 {
+		return nil
+	}
+
+	records := fmt.Sprintf("%s\n", strings.Join(rawRecords, "\n"))
+	if _, err := e.Writer.Write([]byte(records)); err != nil {
+		return pkgerrors.WithStack(err)
+	}
+
```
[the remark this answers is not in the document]
ruiAzevedo19: There is no `Flush` method on the `io.Writer` interface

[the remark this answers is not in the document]
Munsio: Oh right, checkout `WriteEvaluationRecord` you need to create a `csv.NewWriter` with the writer on the `EvaluationFile`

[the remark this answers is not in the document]
ruiAzevedo19: The thing is that the `csv.Write` expects a slice of strings where each string is a field of a record, but the `report` command deals with the entire line:

E.g for the header `model-id,language,task`

- `csv.Writer`: `[]string{"modelA", "golang", "some-task"}` -> `modelA,golang,some-task`
- the report command: `[]string{"modelA,golang,some-task"}` -> "modelA,golang,some-task"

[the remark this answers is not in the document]
Munsio: Alright then instead of reading with a simple reader we could use `https://pkg.go.dev/encoding/csv#Reader.ReadAll`

PR #274, opened 2024-07-18
Always numerize the result path of containerized runs to avoid I/O sync problems

Munsio:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -635,13 +637,12 @@ func (command *Evaluate) evaluateKubernetes(ctx *evaluate.Context) (err error) {
 		}
 
 		// Commands for the evaluation to run inside the container.
+		resultPath := "/var/evaluations/" + log.CleanModelNameForFileSystem(model.ID()) + "_" + strconv.Itoa(i)
```
ruiAzevedo19, line 640: This problem only happens when we pass the same model multiple times on the CLI right?

Munsio: Yes

ruiAzevedo19: I was wondering if we could do it differently so that if there are no duplicate models, we don't need to add a number suffix. We could use a map of visited models, and if a model has already been visited, we then append the number suffix. 

Does it make sense?

ruiAzevedo19 (approved):

PR #287, opened 2024-07-26
Store repositories in JSON

bauersimon:


cmd/eval-dev-quality/cmd/config.go
```diff
@@ -21,6 +24,14 @@ type ModelsConfiguration struct {
 	Available []string
 }
 
+// RepositoryConfiguration holds repository data of how an evaluation was configured.
+type RepositoryConfiguration struct {
+	// Selected holds the repositories selected for an evaluation.
+	Selected []string
```
Munsio, line 30: Is it possible to define which tasks are running for a repository?

Munsio: I guess we should also add the "used" tasks to this list. Then you wouldn't need to cross reference between available/selected if you want to gather metrics.

evaluate/task/repository.go
```diff
@@ -18,13 +18,35 @@ import (
 	"github.com/symflower/eval-dev-quality/util"
 )
 
-// repositoryConfiguration holds the configuration of a repository.
-type repositoryConfiguration struct {
+// RepositoryConfiguration holds the configuration of a repository.
```
Munsio, line 21: Could have been a separate commit, but doesn't matter right now.

evaluate/task/repository.go
```diff
@@ -50,27 +72,21 @@ type Repository struct {
 
 var _ task.Repository = (*Repository)(nil)
 
+// RepositoryConfigurationFileName holds the file name for a repository configuration.
+const RepositoryConfigurationFileName = "repository.json"
```
Munsio, line 76: I would do a check inside `LoadRepositoryConfiguration` to see if the path has the "file" at the end and append it if not, so it would be possible to either define a directory or a file for `LoadRepositoryConfiguration` that way we would have fewer unnecessary exposed variables and fewer `filepath.Join` stuff.

Munsio (changes requested):

cmd/eval-dev-quality/cmd/config.go
bauersimon: We currently don't have a way to select which tasks are running. If a repository supports tasks, they all run. That also makes the current logic easier. I had the version where I stored the tasks and it was just redundant information, should we really add it?

Munsio (approved):

Munsio (approved):

PR #288, opened 2024-07-26
Write openrouter models to CSV and reject models that we want to ignore automatically

bauersimon:


scripts/openrouter-models/main.go
```diff
@@ -47,11 +65,25 @@ MODELS:
...
+		}
+
+		baseModelName := baseModel(model.ID())
+		// Replace existing base model with instruct tuned version.
+		if models[baseModelName] && !isBaseModel(model.ID()) {
+			fmt.Printf("ignoring %q for instruct model %q\n", baseModelName, model.ID())
+
+			delete(models, baseModelName)
+		}
+		models[model.ID()] = true
```
ahumenberger, line 82: I find this a bit complicated. Wouldn't this also work:

```go
if isInstructModel(model.ID()) {
  delete(models, baseModel(model.ID())  // Delete previously added base model.
} else if models[instructModel(model.ID()) {
  continue // Do not add base model if there is already an instruct model.
}

models[model.ID()] = true

```

ahumenberger (changes requested):

scripts/openrouter-models/main.go
```diff
@@ -46,19 +65,37 @@ func main() {
...
+		if !isBaseModel(model.ID()) {
+			baseModelName := baseModel(model.ID())
+			fmt.Printf("(potentially) deleting %q for instruct model %q\n", baseModelName, model.ID())
+
+			delete(models, baseModelName)
+		} else if instructModelName := instructModel(models, model.ID()); instructModelName != "" {
+			fmt.Printf("ignoring %q for instruct model %q\n", model.ID(), instructModelName)
+
+			continue
+		}
```
ahumenberger, line 92: - The `if` below "Skip base model..." is not needed I think
- The "(potentially)" from the print statement can be removed if you do a map lookup first.
- maybe rename "instructModel" to "isInstructModelContained".

ahumenberger (approved):

Issue #282, opened 2024-07-23
Use a JSON configuration file to set up an evaluation run


ruiAzevedo19:
Goal: Allow to export and update a JSON configuration file to run an evaluation. We want to automatically see which models/repositories/tasks are new/gone. This allows us to commit the configuration we are using for a full evaluation for a eval version into the repository.

We want to store:
- available models for providers
- selected models for providers
- available repositories (with their tasks)
- selected repositories (with their tasks)

We want to load (?):
- selected models for providers
- selected repositories

### TODO

**Iteration 1**
- [x] Return all available models by querying the provider's APIs
  - `Provider.Models()`
  -  Providers
     - [x] ~OpenRouter: `https://openrouter.ai/api/v1/models`~ already implemented
     - [x] ~Ollama: `http://127.0.0.1:11434/api/tags`~ already implemented
   - Locally available Ollama models don't say anything about which models are generally available, only which ones are locally available, so ignore Ollama for now (until #283 is in).
 
**Iteration 2**
- [x] Store the available models in JSON file
- [x] Store the selected models in JSON file

**Iteration 3**
- [x] Store the available repositories (with tasks) in JSON file
- [x] Store the selected repositories (with tasks) in JSON file

**Iteration 4**
- [x] Handle JSON file as configuration argument to the evaluation
  - [x] load selected models
  - [x] load selected repositories

**Iteration 5**
- [ ] ~~Also store and load custom provider urls so that they don't need to be carried over manually~~ Follow-up: https://github.com/symflower/eval-dev-quality/issues/307

bauersimon:
@ahumenberger plz check if this makes sense to u :pray:

PR #291, opened 2024-07-29
Load selected models and repositories from the JSON configuration file, to set up an evaluation run

ruiAzevedo19:
Part of #282


ruiAzevedo19:
Based on #287

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -113,6 +115,20 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 	evaluationContext = &evaluate.Context{}
 	evaluationConfiguration = NewEvaluationConfiguration()
 
+	// Load the provided configuration file, if any.
```
bauersimon, line 118: I think we can make this simpler such that we don't have to do all this special handling later on.

- Add a check that `command.Models` and `command.Repositories` must be empty if a configuration is loaded
- just set `command.Models` and `command.Repositories` from the configuration
- all the initialization later will work out of the box (i.e. without having to drag `providedConfiguration` through all the way)

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -113,6 +115,20 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
...
+	var providedConfiguration EvaluationConfiguration
+	if len(command.Configuration) > 0 {
```
bauersimon, line 120: compare with `""` is nicer to read

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -1048,6 +1050,14 @@ func TestEvaluateInitialize(t *testing.T) {
 				}
 			}()
 
+			if tc.Before != nil {
+				tc.Before(t, temporaryDirectory)
+			}
+
+			if len(tc.Command.Configuration) > 0 {
```
bauersimon, line 1057: plz compare with empty string

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -1240,6 +1250,48 @@ func TestEvaluateInitialize(t *testing.T) {
...
+				}
+			`)
+			require.NoError(t, os.WriteFile(filepath.Join(workingDirectory, "config.json"), []byte(configurationContent), 0700))
+		},
+
+		Command: makeValidCommand(func(command *Evaluate) {
+			command.Configuration = "config.json"
+		}),
+
+		ValidateCommand: func(t *testing.T, command *Evaluate) {
```
bauersimon, line 1282: Validate the output configuration itself as well please!

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: you need to convert the repository names back to OS specific ones, cause they will always be UNIX forward slashes

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate_test.go
bauersimon: I.e. in the output config we should again have the models and repositories that we had as inputs.

cmd/eval-dev-quality/cmd/config.go
```diff
@@ -32,6 +34,21 @@ type RepositoryConfiguration struct {
 	Available map[string][]task.Identifier
 }
 
+func (c *RepositoryConfiguration) ConvertNamesToOSSpecific() {
```
bauersimon, line 37: Docstring plz

cmd/eval-dev-quality/cmd/config.go
```diff
@@ -42,6 +59,16 @@ func (c *EvaluationConfiguration) Write(writer io.Writer) error {
 	return nil
 }
 
+// ReadEvaluationConfiguration reads an evaluation configuration file.
+func ReadEvaluationConfiguration(reader io.Reader) (configuration EvaluationConfiguration, err error) {
```
bauersimon, line 63: The configuration holds a lot of data, plus it might be more in the future. Please return a pointer so we avoid copying all of it...

bauersimon, line 63: Now that I am thinking of it... Can you please make `ConvertNamesToOSSpecific` unexported and add it after the reading so it happens automatically. And then move the other conversion (where we change all of them to use `/`) into the storing logic. That way someone using the configuration does not have to know that we internally always use `/`.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -113,6 +115,27 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 	evaluationContext = &evaluate.Context{}
 	evaluationConfiguration = NewEvaluationConfiguration()
 
+	// Load the provided configuration file, if any.
+	if command.Configuration != "" {
+		if command.Models != nil || command.Repositories != nil {
+			command.logger.Panicf("do not provide models and repositories when loading a configuration file")
+		} else {
```
bauersimon, line 122: No need for the `else` block cause the `if` block panics.

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -138,6 +140,12 @@ func TestEvaluateExecute(t *testing.T) {
 				defer tc.After(t, logger, temporaryPath)
 			}
 
+			// Add the temporary path as prefix to the configuration file path if needed.
+			i := slices.Index(tc.Arguments, "--configuration")
+			if i >= 0 {
+				tc.Arguments[i+1] = filepath.Join(temporaryPath, tc.Arguments[i+1])
+			}
```
bauersimon, line 147:
```go
if i := slices.Index(tc.Arguments, "--configuration"); i >= 0 {
```

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -551,6 +551,7 @@ func (command *Evaluate) evaluateDocker(ctx *evaluate.Context) (err error) {
 		"result-path",
 		"runtime-image",
 		"runtime",
+		"configuration",
```
bauersimon, line 554: yes

bauersimon (approved):

PR #292, opened 2024-07-29
Satisfy Kubernetes convention to only have alphanumeric characters and "-" in job names

Munsio:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -663,6 +664,9 @@ func (command *Evaluate) evaluateDocker(ctx *evaluate.Context) (err error) {
 
 // evaluateKubernetes executes the evaluation for each model inside a kubernetes run container.
 func (command *Evaluate) evaluateKubernetes(ctx *evaluate.Context) (err error) {
+	// Define a regex to replace all non alphanumeric characters and "-".
+	kubeNameRegex := regexp.MustCompile(`[^a-zA-Z0-9 -]+`)
```
bauersimon, line 668: I think we don't want the whitespace in the regex.

```sh
[^a-zA-Z0-9 -]+
[^a-zA-Z0-9-]+
```

Munsio: good catch

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -746,6 +746,7 @@ func (command *Evaluate) evaluateKubernetes(ctx *evaluate.Context) (err error) {
 				Command: []string{
 					"kubectl",
 					"wait",
+					"--timeout", "24h",
```
bauersimon, line 749: Is that the whole evaluation or just a single container? I think a single container might be faster... But what if the full evaluation is longer? Would be a bit sad for someone running a huge eval and then it gets killed.

Munsio: It is a single container

Issue #296, opened 2024-07-30
Data visualization based on evaluation CSV files


ruiAzevedo19:
Goal: create a HTML report with graphs for data visualization. 
Tool: [D3.js](https://d3js.org/) library for data visualization graphs

#### TODO

- [x] Create a table for the evaluation CSV file
- [ ] Scatter plot
  - [x] Create a CSV file that stores the models meta information such as pricing and human-readable names
     - [x] Extend the model's interface with a `MetaInformation` function that returns a model's meta information
     - [x] Write the meta information in a CSV file
  - [ ] Create a scatter plot with the model costs and score 

- [ ] error bars over multiple runs, to show variance

bauersimon:
developed internally

PR #298, opened 2024-07-30
Store models meta information in a CSV file, so it can be further used in data visualization

ruiAzevedo19:
Part of #296


provider/openrouter/openrouter.go
```diff
@@ -81,16 +85,49 @@ type Pricing struct {
 
 // Models returns which models are available to be queried via this provider.
```
bauersimon, line 87: I guess we never removed the pricing information from the previous approach. Can you please convert the numeric pricing values to floating point right away? https://stackoverflow.com/questions/9452897/how-to-decode-json-with-type-convert-from-string-to-float64

util/io.go
```diff
@@ -45,3 +45,21 @@ func UniqueDirectory(path string) (uniquePath string, err error) {
 		return true, nil
 	})
 }
+
+// CreateFileIfNotExists creates a file if it does not already exists.
+func CreateFileIfNotExists(filePath string) (file *os.File, err error) {
```
bauersimon, line 50: I think https://pkg.go.dev/os#OpenFile does that already... now that I think about it.

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -80,6 +81,28 @@ func (command *Report) Execute(args []string) (err error) {
 		command.logger.Panicf("ERROR: %s", err)
 	}
 
+	// Create a CSV file that holds the models total scores along with the corresponding model costs.
```
bauersimon, line 84: I guess I would really just create a csv with all the meta information. Not already prepare the required data points for the scatter plot. Basically the JS part gets the data in the most bare bone form and then needs to i.e. match that model `A` has score `1234` (from `evaluation.csv`) and cost `0.5` (from `meta.csv`). Cause then we will have that logic ONLY in the JS part. Right now the Go part already knows that the JS part needs cost vs. score for a scatter plot and needs to prepare the data accordingly.

bauersimon (changes requested):

provider/openrouter/openrouter.go
```diff
@@ -81,16 +85,49 @@ type Pricing struct {
...
+	models = make([]model.Model, len(responseModels.Models))
+	for i, model := range responseModels.Models {
+		models[i] = llm.NewModel(p, p.ID()+provider.ProviderModelSeparator+model.ID)
+	}
+
+	return models, nil
+}
+
+// fetchModels returns the list of models of the provider.
+func (p *Provider) fetchModels() (models ModelsList, err error) {
```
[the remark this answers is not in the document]
ruiAzevedo19: Sounds good

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -80,6 +81,30 @@ func (command *Report) Execute(args []string) (err error) {
...
 	}
 
+	// Create a CSV file that holds the models meta information.
+	var modelsMetaInformationCSVFile *os.File
+	if modelsMetaInformationCSVFile, err = os.OpenFile(filepath.Join(command.ResultPath, "meta.csv"), os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0755); err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+	defer modelsMetaInformationCSVFile.Close()
+
+	// Fetch all models meta information.
```
bauersimon, line 91: Docstring that currently only openrouter supports querying meta information. Maybe even make this more generic by  just going through all providers right away? But also that can happen in the future.

bauersimon (changes requested):

provider/openrouter/openrouter.go
```diff
@@ -53,44 +57,54 @@ func (p *Provider) ID() (id string) {
...
+	if err != nil {
+		return nil, err
+	}
 
-	// Pricing holds the pricing information of a model.
-	Pricing Pricing `json:"pricing"`
-}
+	models = make([]model.Model, len(responseModels.Models))
+	for i, model := range responseModels.Models {
+		models[i] = llm.NewModel(p, p.ID()+provider.ProviderModelSeparator+model.ID, &model)
```
bauersimon, line 72: Found an interesting bug here... You pass the pointer to `model` to the constructor because the model here is a struct. In the next iteration the next model information is copied exactly into that struct... So basically you give every LLM constructor the same pointer cause `model` is never re-allocated. And in the end they all share the meta information from the last model. Try it out ^^.

bauersimon: Possible fix:

```go
for i, model := range responseModels.Models {
	information := model // Reallocate the information cause "model" is overwritten in the next iteration.
	models[i] = llm.NewModel(p, p.ID()+provider.ProviderModelSeparator+model.ID, &information)
}

```

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -87,6 +81,30 @@ func (command *Report) Execute(args []string) (err error) {
...
+	provider := openrouter.NewProvider().(*openrouter.Provider)
+	models, err := provider.Models()
+	if err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+	var modelsMetaInformation []*model.MetaInformation
+	for _, model := range models {
+		modelsMetaInformation = append(modelsMetaInformation, model.MetaInformation())
```
bauersimon, line 99: the provider prefix should be added to the model id within the meta information such that the mapping is correct... maybe we even do this the moment we query the model?

provider/openrouter/openrouter.go
ruiAzevedo19: Good catch

cmd/eval-dev-quality/cmd/report.go
ruiAzevedo19: Updated the docstring. I think we can just query the openrouter for now since it is the only one that has meta information.

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -80,6 +81,30 @@ func (command *Report) Execute(args []string) (err error) {
 		command.logger.Panicf("ERROR: %s", err)
 	}
 
+	// Create a CSV file that holds the models meta information.
+	var modelsMetaInformationCSVFile *os.File
+	if modelsMetaInformationCSVFile, err = os.OpenFile(filepath.Join(command.ResultPath, "meta.csv"), os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0755); err != nil {
```
bauersimon, line 86:
```go
modelsMetaInformationCSVFile, err := ... 
if err := nil { ...
```

saves the explicit type

cmd/eval-dev-quality/cmd/report.go
```diff
@@ -80,6 +81,30 @@ func (command *Report) Execute(args []string) (err error) {
...
+	models, err := provider.Models()
+	if err != nil {
+		command.logger.Panicf("ERROR: %s", err)
+	}
+	var modelsMetaInformation []*model.MetaInformation
+	for _, model := range models {
+		modelsMetaInformation = append(modelsMetaInformation, model.MetaInformation())
+	}
+	metaInformationRecords := report.MetaInformationRecords(modelsMetaInformation)
+	report.SortEvaluationRecords(metaInformationRecords)
```
bauersimon, line 102: Do the sorting within the function where the records are constructed.

I think you can even use a library:

```go
slices.SortFunc(metaInformationRecords, func(a,b []string){
  return strings.Compare(a[0], b[0])
})
```

evaluate/report/csv.go
```diff
@@ -140,6 +140,40 @@ func assessmentFromRecord(assessmentFields []string) (assessments metrics.Assess
...
+	if err := csv.Write(header); err != nil {
+		return pkgerrors.WithStack(err)
+	}
+	if err := csv.WriteAll(records); err != nil {
+		return pkgerrors.WithStack(err)
+	}
+	csv.Flush()
+
+	return nil
+}
```
bauersimon, line 175: Sorry for realizing now but this can be abstract:

`WriteCSV(writer, records, header)`

Maybe you can reuse it somewhere else as well?!

cmd/eval-dev-quality/cmd/report_test.go
```diff
@@ -170,6 +170,7 @@ func TestReportExecute(t *testing.T) {
 				expectedContent := fmt.Sprintf("%s\n%s", strings.Join(report.EvaluationHeader(), ","), claudeEvaluationCSVFileContent)
 				assert.Equal(t, expectedContent, data)
 			},
+			filepath.Join("result-directory", "meta.csv"): nil,
```
bauersimon, line 173: Please check that there are at least 3 lines and all of them differ ^^

bauersimon (changes requested):

cmd/eval-dev-quality/cmd/report.go
ruiAzevedo19: The `SortEvaluationRecords` function sorts records in a generic way, so I would still use that function

ruiAzevedo19: Maybe rename it to `SortRecords`

evaluate/report/csv.go
ruiAzevedo19: We cannot reuse it for the `evaluation.csv` because writing the header and the records happen in two different stages.

bauersimon (approved):

bauersimon:
LGTM... will test-drive once more locally and then merge after #308 is in.

PR #299, opened 2024-07-30
HTML report for data visualization

ruiAzevedo19:
Part of #296


zimmski:
As discussed going intern with all graphing tools.

Issue #300, opened 2024-07-31
Ruby support


ahumenberger:
We want to add language support for Ruby.

- [x] Install Ruby in Docker containers
- [x]  Introduce Ruby as a new language
- [x] Register the Ruby language #327 
- [x] Implement  the `ExecuteTests` method in the Ruby language #327 
- [x] Implement _write-tests_ task
  - [x] Create "plain" repository for Ruby
  - [x] Translate the _light_ repository using a LLM
  - [x] Support Ruby in `symflower test`
- [x] Implement _code_repair_ task
  - [x] Translate the  _mistakes_ repository using a LLM
  - [x] Implement the `Mistakes` method in the Ruby language
  - [x] Add missing tests #327
- [x] Implement _transpile_ task
  - The transpile repository used examples from the light repository, so no need to translate them again
  - [x] Change the way we get the origin/destination languages for transpilation
    - [x] The transpile package names must be e.g. for Go   `transpile-java`, `transpile-ruby`, instead of just `transpile` so it is clear which language we are transpiling from
  - [x] Add missing tests #327
- [x] Documentation
  - [x] Update  documentation with new Ruby support #327 
  - [x] New features #327
    - [x] Supported providers
    - [x] New Reward points
- [x] Merge fixes #327  

#### Follow-up
- [x] collect all mistakes and remove duplicates: https://github.com/symflower/eval-dev-quality/pull/326#discussion_r1732496767

bauersimon:
For new transpilation tasks, there needs to be some small changes to how the eval figures out which languages are targeted.
Cause right now if we have Go as transpilation source, it automatically assumes Java must be the target cause no other language existed yet.

ruiAzevedo19:
### Checking for errors on the Ruby `mistakes` repository

For Java and Go we check for compilation errors with:
- Go: `go build`
- Java: `mvn compile`

Since Ruby is an interpreted language, there is no compilation step. I tried the following approaches.

---

Check Ruby's syntax with `ruby -c lib/<ruby-file>` (see [Command line options](https://ruby-doc.org/docs/ruby-doc-bundle/Manual/man-1.4/options.html))
- argumentsMissing: Syntax OK
- endKeywordMissing: `lib/end_keyword_missing.rb:9: syntax error, unexpected end-of-input, expecting keyword_end`
- importMissing: SyntaxOK
- typeUnknown: SyntaxOK
- variableUnknown: SyntaxOK

Conclusion: only only error caught

--- 

[...]

PR #306, opened 2024-08-01
Check if the repository for the transpile task is valid before running the evaluation, so it is checked just once

ruiAzevedo19:
Part of #263


evaluate/task/task-transpile.go
```diff
@@ -191,3 +187,44 @@ func (t *TaskTranspile) unpackTranspilerPackage(ctx evaltask.Context, fileLogger
...
+// validateTranspileRepository checks if the repository for the "transpile" task is well-formed.
+func validateTranspileRepository(logger *log.Logger, repositoryPath string, destinationLanguage language.Language) (err error) {
+	logger.Printf("validating repository %q", repositoryPath)
+
+	var originLanguage language.Language
+	if _, ok := destinationLanguage.(*golang.Language); ok {
+		originLanguage = &java.Language{}
+	} else {
+		originLanguage = &golang.Language{}
+	}
```
bauersimon, line 200: Can you please already open the Ruby support issue and note that these switches need to be adapted (we do this somewhere else as well AFAIK)?

bauersimon (approved):

evaluate/task/task-transpile.go
bauersimon: Ah the ruby support issue exists already. Will put it there.

ruiAzevedo19: We also do it in the `Run` method.

ruiAzevedo19: Added to https://github.com/symflower/eval-dev-quality/issues/300

PR #308, opened 2024-08-01
Copying docker results did not work without setting the `result-path` parameter

Munsio:


Munsio:
@bauersimon This is a tricky one. Our tests didn't catch it because we are always using a temporary path to store the results in and it only happend when using the default `result-path` where it generates a result directory based on the timestamp.

Converting the path to an absolute one fixes everything but still I am unsure how to add a test-case for it.

bauersimon:
YAAAS... found a reproducer in the CI https://github.com/symflower/eval-dev-quality/actions/runs/10213568583/job/28259341443?pr=310

ruiAzevedo19 (approved):

bauersimon (approved):

PR #309, opened 2024-08-01
Add Ruby as Dependency to the Docker container

Munsio:
Part of #300

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Builder image.
-FROM golang:latest as builder
+FROM golang:latest AS builder
 
 WORKDIR /app
 COPY ./ ./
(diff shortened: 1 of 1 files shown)


Munsio:
Having to build ruby from source increases the Docker build time from ~2:30 to >8:00 minutes.

We should check if it is possible to use prebuilt binaries

docs/reports/v0.6.0-test/ollama_granite-code_8b-instruct-q8_0/README.md
bauersimon, line 1: errrhmmm.... not sure this commit belongs to this branch?

Dockerfile
```diff
@@ -20,6 +32,18 @@ COPY --chown=ubuntu:ubuntu ./Makefile ./Makefile
...
 RUN mkdir -p /app/evaluation
 
+# Install Ruby
+RUN mkdir -p /tmp/compile
+RUN wget https://cache.ruby-lang.org/pub/ruby/3.3/ruby-3.3.4.tar.gz && \
+	tar -xf ruby-3.3.4.tar.gz -C /tmp/compile/ && \
+	rm ruby-3.3.4.tar.gz
+WORKDIR /tmp/compile/ruby-3.3.4
+RUN ./configure --prefix /app/.eval-dev-quality/ruby-3.3.4 --disable-install-doc
+RUN	make install
```
bauersimon, line 42: And that now takes 8 Minutes, right? :see_no_evil: no prebuilt ruby... like ubuntu package? Guess that one is outdated.

Munsio: Added the points into their own Issue so we do keep track: https://github.com/symflower/eval-dev-quality/issues/319

bauersimon (approved):

Dockerfile
Munsio: Yeah artifacts would be stuff you want to pass from one Job to the next in the CI

PR #317, opened 2024-08-06
Infer the language we want to transpile from the file extension, so other languages are supported for transpilation

ruiAzevedo19:
Part of #300

### DIFF
--- a/evaluate/task/task-transpile.go
+++ b/evaluate/task/task-transpile.go
@@ -11,11 +11,10 @@ import (
 	pkgerrors "github.com/pkg/errors"
 	"github.com/symflower/eval-dev-quality/evaluate/metrics"
 	"github.com/symflower/eval-dev-quality/language"
-	"github.com/symflower/eval-dev-quality/language/golang"
-	"github.com/symflower/eval-dev-quality/language/java"
 	"github.com/symflower/eval-dev-quality/log"
 	"github.com/symflower/eval-dev-quality/model"
 	evaltask "github.com/symflower/eval-dev-quality/task"
+	"golang.org/x/exp/maps"
 )
 
 // TaskTranspile holds the transpilation task.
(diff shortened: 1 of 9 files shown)


evaluate/task/task-transpile_test.go
```diff
@@ -546,3 +476,114 @@ func TestValidateTranspileRepository(t *testing.T) {
 		})
 	})
 }
+
+func TestValidateImplementationDirectory(t *testing.T) {
```
bauersimon, line 481: Bit confusing that we now have both `TestValidateTranspileRepository` and `TestValidateImplementationDirectory`. Can you add a docstring to both that the validation is split for simplicity and hence also the tests are split?

evaluate/task/task-transpile_test.go
```diff
@@ -546,3 +476,114 @@ func TestValidateTranspileRepository(t *testing.T) {
...
+
+		ExpectedErrContains: func(err error) {
+			assert.ErrorContains(t, err, "the implementation subdirectories names must be in the form \"implementation-<language>\", but found \"malformed-directory-name\"")
+		},
+	})
+	validate(t, &testCase{
+		Name: "Unsupported language",
+
+		Before: func(implementationDirectory string) {
+			require.NoError(t, osutil.MkdirAll(filepath.Join(implementationDirectory, "implementation-haskell")))
```
bauersimon, line 543: Plz call it `implementation-someunsupportedlanguage` cause hopefully there will never be `someunsupportedlanguage` but once we support `haskell` then this test will fail.

evaluate/task/task-transpile.go
```diff
@@ -11,8 +11,6 @@ import (
 	pkgerrors "github.com/pkg/errors"
```
bauersimon, line 11: would also merge commit with previous

evaluate/task/task-transpile.go
```diff
@@ -170,22 +162,38 @@ func (t *TaskTranspile) Run(ctx evaltask.Context) (repositoryAssessment map[eval
...
+		for _, l := range language.Languages {
+			if languageID == l.ID() {
+				originLanguage = l
+			}
+		}
+		files, err := originLanguage.Files(fileLogger, transpilationDirectoryPath)
+		if err != nil {
+			return nil, "", err
+		}
+		newOriginLanguage := originLanguage
```
bauersimon, line 187: Why the re-pointer. Since the language is already a pointer to begin with, don't think you need it.

bauersimon (changes requested):

evaluate/task/task-transpile_test.go
ruiAzevedo19: Since the `implementation` folder structure changed (it only contains source files now), I removed the function that validates the `implementation` folder. 

Obsolete

evaluate/task/task-transpile.go
```diff
@@ -228,3 +216,48 @@ func validateTranspileRepository(logger *log.Logger, repositoryPath string, dest
...
+			return pkgerrors.Errorf("the implementation directory %q must contain only subdirectories with source code files to transpile, but found one file: %q", implementationDirectoryPath, implementationDirectory.Name())
+		}
+
+		// Check if the implementation folder is well-formed.
+		names := strings.Split(implementationDirectory.Name(), "-")
+		if len(names) != 2 {
+			return pkgerrors.Errorf("the implementation subdirectories names must be in the form %q, but found %q", "implementation-<language>", implementationDirectory.Name())
+		} else if names[0] != "implementation" {
+			return pkgerrors.Errorf("the implementation subdirectories names must be in the form %q, but found %q", "implementation-<language>", implementationDirectory.Name())
+		}
```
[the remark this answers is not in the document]
ruiAzevedo19: Obsolete

testdata/java/transpile/binarySearch/implementation/implementation-golang/binarySearch.go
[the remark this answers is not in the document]
ruiAzevedo19: It makes sense, changing it

evaluate/task/task-transpile.go
```diff
@@ -170,49 +162,66 @@ func (t *TaskTranspile) Run(ctx evaltask.Context) (repositoryAssessment map[eval
...
+		return nil, "", pkgerrors.WithStack(err)
+	}
+
+	for _, file := range files {
+		var originLanguage language.Language
+		for _, l := range language.Languages {
+			if filepath.Ext(file.Name()) == l.DefaultFileExtension() {
+				originLanguage = l
+			}
+		}
```
bauersimon, line 181: Add a comment that we always assume that we only encounter files of languages that we actually support. Because otherwise an error is raised during repository validation.

evaluate/task/task-transpile.go
```diff
@@ -170,49 +162,66 @@ func (t *TaskTranspile) Run(ctx evaltask.Context) (repositoryAssessment map[eval
 	return repositoryAssessment, problems, nil
 }
 
-// unpackTranspilerPackage checks if the testdata repository for the transpilation task is well-formed and returns the path to the implementation file and also the path to the file that holds the stub.
-func (t *TaskTranspile) unpackTranspilerPackage(ctx evaltask.Context, fileLogger *log.Logger, originLanguage language.Language, packagePath string) (originFilePath string, stubFilePath string, err error) {
+// unpackTranspilerPackage returns the path to the implementation file and also the path to the file that holds the stub.
+func (t *TaskTranspile) unpackTranspilerPackage(ctx evaltask.Context, fileLogger *log.Logger, packagePath string) (originFilePathsWithLanguage map[string]language.Language, stubFilePath string, err error) {
```
bauersimon, line 166: Do you remember why the logger is called `fileLogger` here? We usually just call the logger `log`, right?

evaluate/task/task-transpile.go
```diff
@@ -170,49 +162,66 @@ func (t *TaskTranspile) Run(ctx evaltask.Context) (repositoryAssessment map[eval
...
+	}
+
+	for _, file := range files {
+		var originLanguage language.Language
+		for _, l := range language.Languages {
+			if filepath.Ext(file.Name()) == l.DefaultFileExtension() {
+				originLanguage = l
+			}
+		}
+		originFilePathsWithLanguage[filepath.Join(filepath.Join(packagePathAbsolute, "implementation", file.Name()))] = originLanguage
```
bauersimon, line 182: Duplicated `filepath.Join`.

evaluate/task/task-transpile.go
```diff
@@ -170,49 +162,66 @@ func (t *TaskTranspile) Run(ctx evaltask.Context) (repositoryAssessment map[eval
...
-			return pkgerrors.Errorf("package %q must have an \"implementation\" directory with only a %s source file, but found a test file %q", packagePath, originLanguage.Name(), files[0])
+		}
+
+		// Check if the implementation directory is well-formed.
+		for _, file := range files {
+			if file.IsDir() {
+				return pkgerrors.Errorf("the implementation directory %q must contain only source code files to transpile, but found one directory: %q", implementationDirectoryPath, file.Name())
+			}
+			var originLanguage language.Language
+			for _, language := range language.Languages {
```
bauersimon, line 215: Additional validation check: for each language there can only be one implementation file.

evaluate/task/task-transpile_test.go
```diff
@@ -546,3 +526,87 @@ func TestValidateTranspileRepository(t *testing.T) {
...
+			ctx := evaltask.Context{
+				Language:   tc.DestinationLanguage,
+				Repository: repository,
+			}
+			actualOriginFilePathsWithLanguage, actualStubFilePath, actualErr := taskTranspile.unpackTranspilerPackage(ctx, logger, tc.PackagePath)
+			require.NoError(t, actualErr)
+
+			expectedOriginFilePathsWithLanguage := map[string]language.Language{}
+			for filePath, language := range tc.ExpectedOriginFilePathsWithLanguage {
+				expectedOriginFilePathsWithLanguage[filepath.Join(repository.DataPath(), filePath)] = language
```
bauersimon, line 575: I guess in the implementation it does not matter cause the paths are absolute anyways. But here I noticed in the test cases below, that the `ExpectedOriginFilePath`s contain the package in the front, and the stub file path does not. Just a minor inconsistency but a bit confusing. If we change this to `filepath.Join(repository.DataPath(), tc.PackagePath, filePath)`, that should fix it.

evaluate/task/task-transpile.go
bauersimon, line 181: Nitpick: Rather than iterating over the languages over and over again, you could just make a map from default file extension to language. This could even be part of the `language` package, since you need this in the validation as well.

bauersimon (changes requested):

evaluate/task/task-transpile.go
ruiAzevedo19: I don't remember why, I usually call it `logger` AFAIK. I'll change it.

language/language.go
```diff
@@ -53,14 +53,21 @@ type Language interface {
 // Languages holds a register of all languages.
 var Languages = map[string]Language{}
 
+// LanguagesByFileExtension holds the a language file extension and the corresponding language.
```
bauersimon, line 56: docstring typo... would call it `LanguageByFileExtension holds the language for a default file extension`.

evaluate/task/task-transpile.go
```diff
@@ -170,49 +162,69 @@ func (t *TaskTranspile) Run(ctx evaltask.Context) (repositoryAssessment map[eval
...
+
+			originLanguage, ok := language.LanguagesByFileExtension[filepath.Ext(file.Name())]
+			if !ok {
+				return pkgerrors.Errorf("the language extension %q is not supported", filepath.Ext(file.Name()))
+			}
+			if strings.HasSuffix(file.Name(), originLanguage.DefaultTestFileSuffix()) {
+				return pkgerrors.Errorf("the implementation directory %q must contain source files, but found a test file %q", implementationDirectoryPath, files[0])
+			}
+			// Check if the implementation directory only contains one source file per language.
+			originLanguageFiles, err := originLanguage.Files(logger, implementationDirectoryPath)
```
bauersimon, line 221: Nitpick: rather than getting all other files every time, could also just add a `map[language.Language]bool` and throw an error in case we encounter a file where the language is already in the map.

bauersimon (changes requested):

bauersimon (dismissed):

bauersimon (dismissed):

PR #318, opened 2024-08-07
Transpile repository for Ruby

ruiAzevedo19:
Part of #300

### DIFF
--- a/testdata/ruby/transpile/balancedBrackets/Gemfile
+++ b/testdata/ruby/transpile/balancedBrackets/Gemfile
@@ -1,0 +1,7 @@
+source "https://rubygems.org"
+
+gem 'simplecov', require: false, group: :test
+
+gem "rake", "~> 13.2"
+
+gem "minitest", "~> 5.24"
--- a/testdata/ruby/transpile/balancedBrackets/implementation/BalancedBrackets.java
+++ b/testdata/ruby/transpile/balancedBrackets/implementation/BalancedBrackets.java
@@ -1,0 +1,21 @@
+// Copyright (C) rosettacode.org
+package com.eval;
+
+public class BalancedBrackets {
+	static boolean hasBalancedBrackets(char[] charArray) {
+		int brackets = 0;
+		for (char ch : charArray) {
+			if (ch == '[') {
+				brackets++;
+			} else if (ch == ']') {
+				brackets--;
+			} else {
+				return false; // Non-bracket characters.
+			}
+			if (brackets < 0) { // Closing bracket before opening bracket.
+				return false;
+			}
+		}
+		return brackets == 0;
+	}
+}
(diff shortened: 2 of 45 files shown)


testdata/ruby/transpile/balancedBrackets/Gemfile
```diff
@@ -0,0 +1,7 @@
+source "https://rubygems.org"
```
bauersimon, line 1: Shouldn't this PR then not also update the other Go+Java repositories with Ruby implementations?

testdata/ruby/transpile/balancedBrackets/implementation/BalancedBrackets.java
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) rosettacode.org
```
bauersimon, line 1: These are just becoming duplicated files. We should make them symlinks at some point maybe? Added https://github.com/symflower/eval-dev-quality/issues/325.

bauersimon: Because then I would not have to review these by hand ^^

testdata/ruby/transpile/balancedBrackets/lib/balanced_brackets.rb
```diff
@@ -0,0 +1,2 @@
+def has_balanced_brackets(string)
+end
```
bauersimon, line 2: Some models will surely return a string here that says "YES" :laughing:... Not sure if we should just accept that, or add type hints https://sorbet.org/, or add a comment that says "returns a boolean" or something.

bauersimon: Maybe ask Claude Sonnet what is best :stuck_out_tongue_closed_eyes:

bauersimon (changes requested):

testdata/ruby/transpile/balancedBrackets/lib/balanced_brackets.rb
ruiAzevedo19: I asked Claude Sonnet if using a comment it would be enough or should we go with the `Sorbet` approach. The summarized response:

- using comments
  - it's simple and can work well for many LLMs
  - it doesn't require additional setup or dependencies
  - however, it relies on the LLM correctly interpreting and following the comment
- using Sorbet 
  - it's more explicit and formal
  - it provides type information for both the parameter and the return value
  - however, it requires setting up Sorbet (quote: "which might be overkill if you're only using it for LLM code generation)
-  YARD-style
   - documentation generation tool for Ruby
   - does not require additional setup
   - it's more explicit than a simple comment
   - it clearly specifies the parameter and return type
   - Example:
  
```ruby
# @param [String] string
# @return [Boolean]
def has_balanced_brackets(string)
end
```

ruiAzevedo19: Going with the Yard-style approach

bauersimon: Awesome suggestion! Please add that to our blog post as well.

bauersimon: Plz adapt the other stub files as well.

bauersimon (approved):

Issue #301, opened 2024-07-31
Roadmap for v1.0.0


bauersimon:
# Highlights :star2: 

- Spring Boot :leafy_green: unit test support: evaluating how models generate tests for Spring
- Code Migration :repeat: task: asking models to migrate i.e. tests from JUnit 4 to JUnit 5

# Changes :bulb:  

- [x] Development & Management 🛠️
  - [x] Linters in development environment and CI #18 #355 #359 #377 @zimmski 
  - [x] Bump golang.org/x/image from 0.11.0 to 0.18.0 #340 #@dependabot
  - [x] VSCode recommended extensions and debugging configuration #361 @bauersimon 
  - [x] Enable Symflower linter by default in VSC #362 @ruiAzevedo19 
  - [x] Script for tagging and releasing #344 #391 @bauersimon 
  - [x] Add Symflower License in GitHub Actions so the CI works again #389 #393 @zimmski 
  - [x] Clean up old container images #394 #395 @zimmski
  - [x] Tag container images of releases #396 @zimmski
- [x] Documentation 📚
  - [x] Add issue templates for bug and feature requests #338 #342 #345 #346 @bauersimon 
  - [x] Improve wording of templates and documentation #343 @bauersimon 
  - [x] Update symflower.com with DevQualityEval entries #339 #348 @zimmski 
  - [x] Roadmap issue template clarification on workflow and payment page #388 @bauersimon 
  - [x] Roadmap issue template clarification on how to use the roadmap issue itself to create release notes #397
- [x] Evaluation ⏱️
  - [x] refactor, Clarify wording, logs, debugging regarding symflower fix #352 @bauersimon 
  - [x] refactor, Simplify assessment cleanup for tests #353 @ruiAzevedo19 
  - [x] Cleanup logging and log to JSON for easier parsing #379 @bauersimon 
  - [x] fix, Solving Plain once (with or without template) is enough to not get disqualified #360 @bauersimon 
- [x] Models 🤖
  - none
- [x] Reports & Metrics 🗒️
  - [x] Write run count to CSV report #337 @bauersimon  
  - [x] Remove category SVG from the report, because we are moving to a leaderboard #356 @zimmski 
  - [x] Remove categorization and total score of reporting, and the "report" command #357 @zimmski 
  - [x] Remove scoring from within the evaluation and remove and lint unused functions #369 @bauersimon 
  - [x] Remove assessment collection #378 @bauersimon 
[...]

PR #324, opened 2024-08-20
v0.6 results

Munsio:


scripts/evaluation/v0.6-openrouter.sh
```diff
@@ -0,0 +1,77 @@
...
+	--model openrouter/qwen/qwen-2-7b-instruct \
+	--model openrouter/qwen/qwen-72b-chat \
+	--model openrouter/recursal/eagle-7b \
+	--model openrouter/recursal/rwkv-5-3b-ai-town \
+	--model openrouter/rwkv/rwkv-5-world-3b \
+	--model openrouter/teknium/openhermes-2.5-mistral-7b \
+	--model openrouter/togethercomputer/stripedhyena-nous-7b \
+	--model openrouter/undi95/toppy-m-7b \
+	--model openrouter/xwin-lm/xwin-lm-70b \
+	--result-path ./docs/reports/v0.6
```
Munsio, line 77: move that one above the models

scripts/evaluation/v0.6-openrouter.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
+make install
+
+eval-dev-quality evaluate \
+	--runtime docker \
+	--runtime-image ghcr.io/symflower/eval-dev-quality:197ece8 \
```
Munsio, line 7: We should extract the hash into a variable making it more obvious when copying the script for the next run.

scripts/evaluation/v0.6-openrouter.sh
Munsio: Would also introduce a variable for the path making it more obvious when copying it for the next run.

docs/reports/v0.6/README.md
```diff
@@ -0,0 +1,3 @@
+This evaluation ran over several days and is actually assembled from several sub-evaluations. Hence, we only provide the `csv` results for now because the logs are distributed all over.
```
Munsio, line 1: Okay so instead of copying all the results we are only going to publish the CSVs.

Would be nice to have the `report` command doing all this by configuring a source/dest path. Probably a new Issue?

scripts/evaluation/stats.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
Munsio, line 1: This is also something the `report` script should handle automtically. Would add that to the Issue.

scripts/max-score/run.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Usage: script.sh <evaluation-directory>
```
Munsio, line 3: This should too be added to the Issue

Munsio (commented):
Does look good. About the results I would like to have @zimmski check those too again if they are correct.

docs/reports/v0.6/README.md
bauersimon: Added to https://github.com/symflower/eval-dev-quality/issues/301

scripts/evaluation/stats.sh
bauersimon: Added to https://github.com/symflower/eval-dev-quality/issues/301

scripts/max-score/run.sh
bauersimon: Added to https://github.com/symflower/eval-dev-quality/issues/301

bauersimon (approved):

PR #326, opened 2024-08-26
Define the mistakes logic for Ruby

ruiAzevedo19:


language/ruby/language.go
```diff
@@ -62,9 +66,35 @@ func (l *Language) ExecuteTests(logger *log.Logger, repositoryPath string) (test
...
+		},
+
+		Directory: repositoryPath,
+	})
+	if err != nil {
+		if output != "" {
+			return extractMistakes(output), nil
+		}
+
+		return nil, pkgerrors.Wrap(err, "no output to extract errors from")
```
bauersimon, line 86: plz swap the logic to handle errors first. i.e. `if output == "" {...`

language/ruby/language.go
```diff
@@ -62,9 +66,35 @@ func (l *Language) ExecuteTests(logger *log.Logger, repositoryPath string) (test
...
 
+var mistakesRe = regexp.MustCompile(`\d\) Error:\n\w+#\w+:\n(.*)\n\s*(.*)`)
+
 // Mistakes builds a Ruby repository and returns the list of mistakes found.
 func (l *Language) Mistakes(logger *log.Logger, repositoryPath string) (mistakes []string, err error) {
-	logger.Panic("not implemented")
+	output, err := util.CommandWithResult(context.Background(), logger, &util.Command{
+		Command: []string{
+			"rake",
+			"test",
```
bauersimon, line 76: Please document that this is sub-optimal since it only works if there are tests for sure. I.e. we might introduce different ruby tasks in the future where just a script is run. So this won't work then anymore.

language/ruby/language.go
```diff
@@ -62,9 +66,45 @@ func (l *Language) ExecuteTests(logger *log.Logger, repositoryPath string) (test
 	return testResult, problems, nil
 }
 
+var mistakesRe = regexp.MustCompile(`\d\) Error:\n\w+#\w+:\n(.*)\n\s*(.*)`)
+
+var mistakesSyntaxErrorRe = regexp.MustCompile(`.* \(SyntaxError\)`)
```
bauersimon, line 71: plz docstring, also can you name the first one more clearly? cause the second one is named more precise as well so from just looking I'm not sure what the difference is

language/testing/language.go
```diff
@@ -85,6 +86,10 @@ func (tc *TestCaseMistakes) Validate(t *testing.T) {
 		actualMistakes, actualErr := tc.Language.Mistakes(logger, repositoryPath)
 		require.NoError(t, actualErr)
 
-		assert.Equal(t, tc.ExpectedMistakes, actualMistakes)
+		if tc.ExpectedMistakesContains != nil {
+			tc.ExpectedMistakesContains(actualMistakes)
+		} else {
+			assert.Equal(t, tc.ExpectedMistakes, actualMistakes)
+		}
```
bauersimon, line 93: Please always hand through the testing object! This can get very confusing if assertions fail. See here:

```go
func TestSomething(tOuter *testing.T) {
    ...
    validate(tOuter, &testCase{
        Name: "FOO"
        SomeFunctionWithLogic: func(tInner *testing.T){
            assert.True(tOuter, false, "messageOuter")
            // If this fails, the error message will show up with just "TestSomething" cause it is the outer T instance.
            
            assert.True(tInner, false, "messageInner")
            // If this fails, the error message will show up with "TestSomething/FOO" and we know exactly where it comes from.
        }
    })
}
```

language/ruby/language_test.go
```diff
@@ -62,3 +64,318 @@ func TestLanguageImportPath(t *testing.T) {
...
+	})
+}
+
+func TestExtractMistakes(t *testing.T) {
+	type testCase struct {
+		Name string
+
+		RawMistakes string
+
+		ExpectedMistakes func(mistakes []string)
```
bauersimon, line 92: Please pass along the `*testing.T` object here as well.

language/ruby/language_test.go
```diff
@@ -62,3 +64,318 @@ func TestLanguageImportPath(t *testing.T) {
...
+			rake aborted!
+			Command failed with status (1)
+
+			Tasks: TOP => test
+			(See full trace by running task with --trace)
+		`,
+
+		ExpectedMistakes: func(mistakes []string) {
+			assert.Contains(t, mistakes[0], "ArgumentError: wrong number of arguments (given 1, expected 0)")
+			assert.Contains(t, mistakes[0], "argument_missing.rb:1:in `argument_missing'")
```
bauersimon, line 143: Since this is the core of how we extract the error messages, I would like to see an actual 1:1 assertion, not just contains. Just to make sure that the extracted message does not contain too much junk.

bauersimon (changes requested):

language/ruby/language.go
```diff
@@ -66,9 +70,47 @@ func (l *Language) ExecuteTests(logger *log.Logger, repositoryPath string) (test
...
+	for _, result := range mistakesErrorRe.FindAllStringSubmatch(rawMistakes, -1) {
+		if !strings.Contains(result[2], "_test.rb") {
+			return []string{result[1] + " : " + result[2]}
+		}
+	}
+
+	// If no errors match the regexp then we check for syntax errors.
+	for _, result := range mistakesSyntaxErrorRe.FindAllStringSubmatch(rawMistakes, -1) {
+		mistakes = append(mistakes, result[0])
+	}
```
bauersimon, line 113: follow up: collect all mistakes (also for the error case) but then make them unique (remove duplicates)

bauersimon (approved):

PR #349, opened 2024-10-02
Script to convert all SVG files to PNG with Inkscape

bauersimon:


scripts/convert_svg_to_png.sh
```diff
@@ -0,0 +1,38 @@
...
+        echo "Renamed $svg_file to $new_svg_name"
+    fi
+    
+    # Generate the PNG filename
+    png_file=$(clean_filename "${svg_file%.svg}").png
+    
+    # Convert SVG to PNG using Inkscape
+    inkscape --export-type=png \
+             --export-filename="$png_file" \
+             --export-width=3600 \
```
Munsio, line 32: Why 3600? Isn't that too big?

zimmski: We want a high-resolution version for the blog posts.

PR #354, opened 2024-10-03
Always apply "symflower fix" and refactor "write test" task and LLM prompting to prepare for templates

bauersimon:


model/llm/llm.go
```diff
@@ -123,14 +129,14 @@ var llmCodeRepairSourceFilePromptTemplate = template.Must(template.New("model-ll
 	- {{.}}{{ end }}
 `)))
 
-// llmCodeRepairSourceFilePrompt returns the prompt to code repair a source file.
-func llmCodeRepairSourceFilePrompt(data *llmCodeRepairSourceFilePromptContext) (message string, err error) {
+// Format returns the prompt to code repair a source file.
+func (ctx *llmCodeRepairSourceFilePromptContext) Format() (message string, err error) {
```
ruiAzevedo19, line 133: Nitpick: Wouldn't make more sense calling this `Prompt`?

bauersimon: I think it is just a context for a template. So I guess format makes sense. Prompt is again rather specific...

PR #356, opened 2024-10-04
Remove category SVG from the report, because we are moving to a leaderboard

zimmski:


bauersimon (commented):
Does the linter not check for unused code? I though i.e. `evaluate/metrics/category.go` would be obsolete with this?

zimmski:
> evaluate/metrics/category.

Still used in the markdown generation.

PR #363, opened 2024-10-14
Spring boot plain write tests

bauersimon:


evaluate/evaluate.go
```diff
@@ -207,7 +209,7 @@ func Evaluate(ctx *Context) (assessments *report.AssessmentStore) {
 		}
 		for _, repositoryPath := range relativeRepositoryPaths {
 			// Do not include "plain" repositories in this step of the evaluation, because they have been checked with the common check before.
-			if !repositoriesLookup[repositoryPath] || strings.HasSuffix(repositoryPath, RepositoryPlainName) {
+			if (len(ctx.RepositoryPaths) > 0 && !repositoriesLookup[repositoryPath]) || filepath.Base(repositoryPath) == RepositoryPlainName {
```
zimmski, line 212: Commit subject: what and why

What did we before, what are we doing now. why.

bauersimon:
Problem here is that models can also score points even if they don't generate Spring Boot tests. I.e. here is a result from PaLM 2 from the test drive:

```java
import com.example.controller.SomeController;
import org.junit.jupiter.api.Test;

import static org.junit.jupiter.api.Assertions.assertEquals;

class SomeControllerTests {

	@Test
	void helloGet() {
		SomeController controller = new SomeController();
		String result = controller.helloGet();
		assertEquals("get!", result);
	}
}
```

Compiles, get's coverage, but IMHO not what the task is. So it should not get points. Just to keep in mind for the next iteration.

---

What we need to solve:

- modify the prompt such that we can tell the model explicitly that we want spring tests
- add validation logic in task such that we can flag cases where some requirements are not met

We could have these values configurable in `repository.json`. I.e.:

```json
// testdata/java/spring-plain/repository.json
{
  "tasks": ["write-tests"],
  "attributes": {
    "test-framework": "JUnit 5 for Spring Boot" // Overwrites the default Java "JUnit 5" test framework we use in the prompt.
  },
  "validation": {
    "execution": [
      "Initializing Spring" // The model only gets points if this (regex) is in the execution log.
    ]
  }
}
```

PR #367, opened 2024-10-24
Support Spring

bauersimon:


task/config.go
```diff
@@ -0,0 +1,65 @@
+package task
```
ruiAzevedo19, line 1: Putting this in `task/config.go`  sounds like this is a task configuration but it is rather a repository configuration right? Should the file be renamed from `config.go` to `repository.go`? Or even put this in a `repository` package e.g. `repository/config.go`?

model/symflower/symflower.go
```diff
@@ -16,34 +16,55 @@ import (
...
+}
+
+// NewModelSymbolicExecution returns a symbolic execution Symflower model.
+func NewModelSymbolicExecution() (model *Model) {
+	return NewModelSymbolicExecutionWithTimeout(defaultSymflowerTimeout)
 }
 
-// NewModel returns a Symflower model.
-func NewModel() (model *Model) {
+// NewModelSymbolicExecutionWithTimeout returns a Symflower model with a given timeout.
```
ruiAzevedo19, line 38: `returns a symbolic execution Symflower model`

model/symflower/symflower.go
```diff
@@ -16,34 +16,55 @@ import (
...
+		symbolicExecutionTimeout: timeout,
+
+		command: "unit-tests",
+		id:      "symbolic-execution",
 	}
 }
 
-// NewModelWithTimeout returns a Symflower model with a given timeout.
-func NewModelWithTimeout(timeout time.Duration) (model *Model) {
+// NewModelSmartTemplate returns a symbolic execution Symflower model.
```
ruiAzevedo19, line 48: This is not a symbolic execution model but rather test templates.

model/symflower/symflower_test.go
```diff
@@ -139,6 +144,27 @@ func TestModelGenerateTestsForFile(t *testing.T) {
 			},
 		})
 	})
+	validate(t, &testCase{
+		Name: "Skeleton",
```
ruiAzevedo19, line 148: Maybe `Smart template`?

README.md
```diff
@@ -226,6 +226,17 @@ Each repository can contain a configuration file `repository.json` in its root d
...
+Depending on the task, it can be beneficial to exclude parts of the repository from explicit evaluation. To give a concrete example: Spring controller tests can never be executed on their own but need a supporting [`Application` class](https://docs.spring.io/spring-boot/reference/testing/spring-boot-applications.html#testing.spring-boot-applications.using-main). But such a file should never be used itself to prompt models for tests. Therefore, it can be excluded through the `repository.json` configuration:
+
+```json
+{
+  "tasks": ["write-tests"],
+  "ignore": ["src/main/java/com/example/Application.java"]
+}
+```
+
+This `ignore` setting is currently only respected for the test generation task `write-tests`.
```
ruiAzevedo19, line 238: `... is currently applicable only for the...` maybe?

task/config.go
bauersimon: The interface for `Repository` is also defined within `task`. That's why I moved it there. So we would probably want to move that one as well. Would leave it for now.

ruiAzevedo19: Let's leave it for now then

ruiAzevedo19 (approved):

PR #371, opened 2024-12-09
Remove evaluation logs for older versions

bauersimon:


ahumenberger:
So you are removing v0.2 - v0.6. I don't know anything about this removal, so I'm hesitant to merge. But if that's what we want, please do.

PR #373, opened 2024-12-10
Update README.md with new "support us" message and call-to-action

ghost:
Use a strong message, don't bury the lede, update to "latest results", add CTA image


bauersimon (approved):
LGTM

PR #378, opened 2025-01-08
Remove assessment collection

bauersimon:


evaluate/report/collection_test.go
```diff
@@ -64,124 +64,132 @@ func TestAssessmentPerModelPerLanguagePerRepositoryWalk(t *testing.T) {
...
 
-func assessmentTuplesToStore(at metricstesting.AssessmentTuples) (store *AssessmentStore) {
+func assessmentTuplesToStore(t *testing.T, at metricstesting.AssessmentTuples) (store *AssessmentStore) {
 	store = NewAssessmentStore()
+	modelMocks := map[string]model.Model{}
+	languageMocks := map[string]language.Language{}
+
 	for _, a := range at {
-		store.Add(a.Model, a.Language, a.RepositoryPath, a.Task, a.Assessment)
+		m := modelMocks[a.Model]
```
ruiAzevedo19: What do you think about using `m, ok := modelMocks[a.Model]` since it is more idiomatic

evaluate/evaluate_test.go
```diff
@@ -575,43 +569,44 @@ func TestEvaluate(t *testing.T) {
 					&metricstesting.AssessmentTuple{
 						Model:          mockedModel.ID(),
 						Language:       languageGolang.ID(),
-						RepositoryPath: repositoryNextPath,
+						RepositoryPath: repositoryPlainPath,
```
ruiAzevedo19, line 572: Why changing the repository under test?

evaluate/report/collection_test.go
bauersimon: Yeah... would be more pretty wouldn't it... :thinking:

ruiAzevedo19: The file is removed in the following commit, so ignoring this

PR #379, opened 2025-01-09
JSON, structured logging

bauersimon:


bauersimon:
todo

- [ ] stdout ALL
- [ ] query responses UUID to connect with the query

log/logger.go
```diff
@@ -356,20 +316,22 @@ var defaultLogFileSpawners = []logFileSpawner{
...
 			repositoryName := attributes[AttributeKeyRepository]
 			taskIdentifier := attributes[AttributeKeyTask]
 
-			return filepath.Join(resultPath, taskIdentifier, CleanModelNameForFileSystem(modelID), languageID, repositoryName, "evaluation.log")
+			file, err := openLogFile(filepath.Join(resultPath, taskIdentifier, CleanModelNameForFileSystem(modelID), languageID, repositoryName, "evaluation.log"))
+			if err != nil {
+				return nil, err
+			}
+
+			return slog.NewJSONHandler(file, nil), nil
```
ahumenberger: Every `handlerSpawner` has now the this redundant block. What about doing this:

```
filePath := filepath.Join(...)

return newJSONHandlerForFile(filePath)
```

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -182,24 +182,24 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 
 		if command.Runtime == "docker" {
 			if _, err := exec.LookPath("docker"); err != nil {
-				command.logger.Panic("docker runtime could not be found")
+				command.logger.Panicf("docker runtime could not be found")
```
ahumenberger, line 185: Not sure what the what the consequences of these changes are. Why is this easier for parsing?

cmd/eval-dev-quality/cmd/evaluate_test.go
```diff
@@ -314,10 +314,10 @@ func TestEvaluateExecute(t *testing.T) {
 					validateREADME(t, data)
 				},
 				filepath.Join("result-directory", string(evaluatetask.IdentifierWriteTests), "symflower_symbolic-execution", "golang", "golang", "plain", "evaluation.log"): func(t *testing.T, filePath, data string) {
-					assert.Contains(t, data, "coverage objects: [{")
+					assert.Contains(t, data, "\"msg\":\"coverage objects\",\"objects\":\"[{")
```
ahumenberger, line 317: commit subject: "Adapting test cases to changes in logging"
And a message body that makes sense to me, please :wink:

log/logger.go
```diff
@@ -358,6 +366,23 @@ func (s handlerSpawner) NeedsSpawn(attributes map[AttributeKey]string) bool {
 	return true
 }
 
+// createFileHandlerForParent creates a JSON handler for logging into a file.
+// REMARK The returned handler may also delegate to any "plainTextHandler", that the parent is already delegating to.
```
ahumenberger, line 338: we only delegate to `plainTextHandler` because STDOUT is such a handler, and because all other handlers are not. is that correct?
if we would delegate to all other handlers as well, we would have the same log file tree structure as before, right?

cmd/eval-dev-quality/cmd/evaluate.go
bauersimon: The `Panicf` does nothing for parsing, just adjusting the signature. As for all the others:

`log.Info("something", "contextA", contextA, "contextB" contextB, ...)`

ends up in the JSON logs as

```json
{
    "msg": "something",
    "contextA": contextA...,
    "contextB": contextB...,
    ...
}
```

which... is nice if you parse the JSON.

log/logger.go
bauersimon: I am introducing a helper later on anyways. So maybe I just stuff it in there?

Issue #403, opened 2025-01-30
Introduce `$binary evaluate ... --run-id-starts-at $integer` as option to start the "run ID" from something other than "1"


zimmski:
### Please help us keep the number of duplicated issues small. We kindly ask you to add your input to the appropriate issue or PR in case your feature idea is already being tracked.

- [x] I have searched the existing [feature requests](https://github.com/symflower/eval-dev-quality/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement), latest [roadmap](https://github.com/symflower/eval-dev-quality/issues?q=is%3Aopen+is%3Aissue+label%3Aroadmap) and [open pull requests](https://github.com/symflower/eval-dev-quality/pulls).

### Problem

When we split up an evaluation in our CI and let multiple CI jobs use the "--run 1" option, we only receive "1" as run IDs.

### Solution (optional)

Tasks:
- [ ] Simplify how runs are created for all our code. It seems that this is right now all over the place.
- [ ] Introduce `--run-id-starts-at $integer`  as option to start the run ID from something else than 1. Must be >0.

A scenario like `$binary --runs 5 --run-id-starts-at 3` should then produce 3 4 5 6 7 as IDs.

bauersimon:
seems to be fixed through https://github.com/symflower/eval-dev-quality/commit/a8995ec389e5624de34a72abeed60be1bb34aae5

PR #412, opened 2025-02-05
CLI flag for specifying a run ID starting index to be able to distribute runs

Munsio:
Part of #403

### DIFF
--- a/cmd/eval-dev-quality/cmd/evaluate.go
+++ b/cmd/eval-dev-quality/cmd/evaluate.go
@@ -73,6 +73,8 @@ type Evaluate struct {
 	Configuration string `long:"configuration" description:"Configuration file to set up an evaluation run."`
 	// ExecutionTimeout holds the timeout for an execution.
 	ExecutionTimeout uint `long:"execution-timeout" description:"Execution timeout for compilation and tests in minutes." default:"5"`
+	// RunIDStartsAt holds the offset increment for the run id used in creating the result folders.
+	RunIDStartsAt uint `long:"run-id-starts-at" description:"Sets the starting index for the run ID." default:"1"`
 	// Runs holds the number of runs to perform.
 	Runs uint `long:"runs" description:"Number of runs to perform." default:"1"`
 	// RunsSequential indicates that interleaved runs are disabled and runs are performed sequentially.
@@ -174,6 +176,8 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 		} else {
 			language.DefaultExecutionTimeout = time.Duration(command.ExecutionTimeout) * time.Minute
 		}
+
+		evaluationContext.RunIDStartsAt = command.RunIDStartsAt
 
 		if command.Runs == 0 {
 			command.logger.Panicf("number of configured runs must be greater than zero")
--- a/evaluate/evaluate.go
+++ b/evaluate/evaluate.go
@@ -38,6 +38,8 @@ type Context struct {
 	// TestdataPath determines the testdata path where all repositories reside grouped by languages.
 	TestdataPath string
 
+	// RunIDStartsAt holds the run ID starting index created when running a evaluation multiple times.
+	RunIDStartsAt uint
 	// Runs holds the number of runs to perform.
 	Runs uint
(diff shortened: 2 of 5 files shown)


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -175,6 +177,11 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
 			language.DefaultExecutionTimeout = time.Duration(command.ExecutionTimeout) * time.Minute
 		}
 
+		if command.RunIDOffset == 0 {
+			command.logger.Panicf("number of run-id-starts-at must be greater than zero")
+		}
```
ahumenberger, line 182: @Munsio Why?

Offset could also be zero for the first run, no?

Munsio: Yeah I just added it so it is there, would have probably found it myself later on.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -73,6 +73,8 @@ type Evaluate struct {
 	Configuration string `long:"configuration" description:"Configuration file to set up an evaluation run."`
 	// ExecutionTimeout holds the timeout for an execution.
 	ExecutionTimeout uint `long:"execution-timeout" description:"Execution timeout for compilation and tests in minutes." default:"5"`
+	// RunIDOffset holds the offset increment for the run id used in creating the result folders.
+	RunIDOffset uint `long:"run-id-starts-at" description:"Sets the offset for the ID used in creating result folders." default:"1"`
```
Munsio, line 77: Should we set the default to "0" so it does not affect anything unless you set it?

evaluate/evaluate.go
```diff
@@ -148,6 +150,7 @@ func Evaluate(ctx *Context) {
 								} else {
 									runCount = rl + 1
 								}
+								runID := ctx.RunIDOffset + runCount
```
Munsio, line 153: If RunIDOffset is by default `1` shouldn't all evaluations where it is not set automatically start with `runID: 2`

cmd/eval-dev-quality/cmd/evaluate.go
ahumenberger: Removing that guard.

cmd/eval-dev-quality/cmd/evaluate.go
ahumenberger: Yeah the default should be zero. The offset is basically the number of runs that are already run by some other job.

evaluate/evaluate.go
ahumenberger: Not sure what you mean here.

cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -73,6 +73,8 @@ type Evaluate struct {
 	Configuration string `long:"configuration" description:"Configuration file to set up an evaluation run."`
 	// ExecutionTimeout holds the timeout for an execution.
 	ExecutionTimeout uint `long:"execution-timeout" description:"Execution timeout for compilation and tests in minutes." default:"5"`
+	// RunIDOffset holds the offset increment for the run id used in creating the result folders.
+	RunIDOffset uint `long:"run-id-starts-at" description:"Sets the offset for the ID used in creating result folders." default:"0"`
```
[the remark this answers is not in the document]
ahumenberger: That depends if the cli option is an "offset" or an "id-starts-at". I'm just working with what I got here.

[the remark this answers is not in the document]
ahumenberger: Turned it into `RundIDStartsAt`

evaluate/evaluate.go
bauersimon: Yeah an offset is usually something that is added. So I also would expect
- offset 0: first run ID 1
- offset 1: first run ID 2
- ...

bauersimon: We call it `run-id-starts-at` in the CLI though which is no offset. So I would keep it.

PR #421, opened 2025-02-19
fix, Allow selecting models with attributes for openRouter as well

ahumenberger:


cmd/eval-dev-quality/cmd/evaluate.go
```diff
@@ -489,10 +490,18 @@ func (command *Evaluate) Initialize(args []string) (evaluationContext *evaluate.
...
+				m, ok = models[modelIDWithProvider]
 				if !ok {
 					command.logger.Panicf("ERROR: model %q does not exist for provider %q. Valid models are: %s", modelIDsWithProviderAndAttributes, providerID, strings.Join(modelIDs, ", "))
 				}
+
+				// If a model with attributes is requested, we add the base model plus attributes as new model to our list.
+				if len(attributes) > 0 {
+					modelWithAttributes := m.Clone()
+					modelWithAttributes.SetAttributes(attributes)
+					models[modelWithAttributes.ID()] = modelWithAttributes
```
Munsio, line 502: This one is unnecessary. All checks on `models[]` have already been done so there is no point in adding it.

Also checks are done with `provider/model` but with the change to `model.ID`  here we are setting it with parameter.

PR #432, opened 2025-03-16
Upgrade Go to 1.24.1 (from 1.23.6) and all Go packages

zimmski:

PR #439, opened 2025-03-26
Bump symflower version to v48562 (from v46727)

Munsio:


Munsio:
With this one in we should also be able to merge https://github.com/symflower/eval-dev-quality/pull/432 and then release a new version.

bauersimon (approved):

Munsio:
Will close this PR as it is incorporated into #432

PR #451, opened 2025-04-24
Format error slices in logs

bauersimon:


evaluate/evaluate.go
```diff
@@ -172,9 +171,6 @@ func Evaluate(ctx *Context) {
 								if err != nil {
 									ps = append(ps, err)
```
Munsio, line 172: Why do we have two loops executing the task?

Why is `problemsPerModel` defined and there are appends on it but it actually never gets used.

Why are we logging the `ps` from this loop on line 178 when the taskLogger should already log them.
Do we really need to return problems from a Task anymore after this change? Or can we actually remove the whole return param?

Munsio (changes requested):

evaluate/evaluate.go
bauersimon: - two loops cause `plain`
- was originally defined when we collected everything and wrote it to a single csv/log file in the end
- this is an extra log to clarify that the model will be disqualified cause it did not solve `plain`

Munsio (dismissed):

evaluate/evaluate.go
```diff
@@ -292,8 +288,7 @@ func Evaluate(ctx *Context) {
 
 									Logger: logger,
 								}
-								assessment, ps, err := task.Run(taskContext)
-								problemsPerModel[modelID] = append(problemsPerModel[modelID], ps...)
+								assessment, _, err := task.Run(taskContext)
```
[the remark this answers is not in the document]
zimmski: Wasn't really relevant which error we saw, but that will get increasingly important. Would be also a nice assessment.

PR #455, opened 2025-05-13
fix, Configuration for pruning tagged container images

ahumenberger:


Munsio:
So we need to add all to prune to keep only the ones we want!?

ahumenberger:
> So we need to add all to prune to keep only the ones we want!?

Yes, because of https://github.com/vlaurin/action-ghcr-prune#:~:text=Disabled%20by%20default%20(ie.%20no%20versions%20pruned%20based%20on%20tags).

Munsio (approved):

PR #456, opened 2025-05-14
Let the "write-tests" prompt ask to extend the given source file in case tests should be next to the source

ahumenberger:


bauersimon (approved):

bauersimon:
LGTM. `test-integration` will fail though

bauersimon (approved):

